PR #2, opened 2015-06-29
Touch support

iamdustan:
Wanted to see how this performed on mobile browsers. Only tested in an iOS simulator so far, but it nailed it.

with-touch

### DIFF
--- a/Demo1.js
+++ b/Demo1.js
@@ -27,6 +27,15 @@ export default React.createClass({
     };
   },
 
+  handleTouchStart: function(key, pressLocation, event) {
+    this.handleMouseDown(key, pressLocation, event.touches[0]);
+  },
+
+  handleTouchMove: function(event) {
+    event.preventDefault();
+    this.handleMouseMove(event.touches[0]);
+  },
+
   handleMouseMove: function({pageX, pageY}) {
     let {order, lastPress, isPressed, delta: [dx, dy]} = this.state;
     if (isPressed) {
@@ -77,6 +86,8 @@ export default React.createClass({
     let {mouse, order, lastPress, isPressed} = this.state;
     return (
       <Springs
+        onTouchMove={this.handleTouchMove}
+        onTouchEnd={this.handleMouseUp}
         onMouseMove={this.handleMouseMove}
         onMouseUp={this.handleMouseUp}
         className="demo1"
@@ -85,6 +96,7 @@ export default React.createClass({
           <div
             key={key}
             onMouseDown={this.handleMouseDown.bind(null, key, [x, y])}
+            onTouchStart={this.handleTouchStart.bind(null, key, [x, y])}
             className="demo1-circle"
             style={{
               backgroundColor: allColors[key],
--- a/Demo2.js
+++ b/Demo2.js
@@ -3,6 +3,12 @@
 import React from 'react';
 import Springs from './Springs';
 
+if (!String.prototype.includes) {
+  String.prototype.includes = function() {'use strict';
+    return String.prototype.indexOf.apply(this, arguments) !== -1;
+  };
+}
+
 export default React.createClass({
   getInitialState: function() {
     return {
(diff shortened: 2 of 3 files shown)


Demo2.js
```diff
@@ -3,6 +3,12 @@
 import React from 'react';
 import Springs from './Springs';
 
+if (!String.prototype.includes) {
+  String.prototype.includes = function() {'use strict';
+    return String.prototype.indexOf.apply(this, arguments) !== -1;
+  };
+}
+
```
chenglou: https://babeljs.io/docs/learn-es2015/#math-number-string-object-apis

Babel polyfills these already, we don't need this =)

Demo1.js
```diff
@@ -27,6 +27,15 @@ export default React.createClass({
     };
   },
 
+  handleTouchStart: function(key, pressLocation, event) {
+    this.handleMouseDown(key, pressLocation, event.touches[0]);
```
chenglou: Hey what's `touches[0]` again? Would it be clearer if we used destructuring?
`{touches: [something]}`

Demo2.js
iamdustan: ah. Does this repo need them enabled then? It was definitely throwing an error for me.

Demo1.js
iamdustan: https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/touches

Maybe to destructuring? Looking at what you were actually using in `MouseDown` and `MouseMove` I suspected that a TouchList touches object would have the same properties and I could get away with it.

It would be fun to try supporting multiple touches on this and seeing what happens :)

chenglou:
Whah =D thanks
Well I sure am glad it doesn't lag for like, 11 balls lol
I allocate like crazy in the current `Springs` so that's good news. We'll have some wiggle room.
If you want, you can try it on React-Native =).

I'll merge after we resolve the comments

Demo2.js
chenglou: Oh sorry, seems I was mistaken. I was testing on chrome mainly and babel-node and thought it was auto enabled. Guess we should just enable the option instead?

Demo1.js
iamdustan: Ah. And `touches[0]` specifically is one of these: https://developer.mozilla.org/en-US/docs/Web/API/Touch

chenglou: What I mean is, maybe instead of `handle(a, b, event.touches[0])`, `handle(a, b, firstPoint)` is clearer in conjunction with `destructuring on `event`: ({touches: [firstPoint]})`? At this point it's just a nit. IF you think the current one's clearer let's merge.

chenglou:
Too much trouble (aka, too lazy) to check how to include the babel runtime. Merge time!

And yeah, secret demo is coming up. It's already here locally =)

Demo1.js
iamdustan: Weeeeeell, basically I haven’t looked at destructuring arguments at all, and definitely not destructuring array arguments inside of objects.

Demo2.js
iamdustan: I can check into this one and open a PR whenever I do.

chenglou:
Update: not 60fps on iPhone 6. Somewhat expected but there's a lot of room for optimization. Maybe I'll try on react-native later and blame it on the dom lol

iamdustan:
Where is that favorite and retweet button? :smile:

chenglou:
SSSHHHHHH I'm not done yet! The repo's supposed to be an open secret.
The plan's probably rolling this into tween-state as the new API. People who want mixins + regular ease functions (why would you now?) can use that. Otherwise you import this Springs component instead. In the future for backward compat we can deprecate `tweenState.Mixin` and turn in into a `<Tween />` component, or simply abstract `Springs` away into that more general `Tween` component.

Also, real `Decay` component coming up...

FoxxMD:
:+1:

PR #3, opened 2015-07-02
Add hot reloading in development

gaearon:
Here's a proof of concept of hot reloading. I'm not sure whether this is useful with stateful spring system, but you can play with it :-)

```
npm install
npm start
open http://localhost:3000/demo1
```

You can also make a production build

```
npm run build
```

and open the regular HTML file.


chenglou:
:+1: :D

PR #1, opened 2015-06-15
Add standard CSS `transform` property

iamdustan:
I didn’t listen to your README warning.

This looks awesome.

PR #6, opened 2015-07-03
- Enabling touchMove support for mobile devices for Demo #1

jevgenig:
Quick "semi-hacky" way to enable Demo#1 for mobile devices


chenglou:
Thanks!

PR #7, opened 2015-07-03
readme code formatting

hzoo:


chenglou:
Sure, thanks! But the readme's gonna be rewritten properly tonight, so let's hold off for these for now.

PR #8, opened 2015-07-03
use method shorthand

hzoo:


chenglou:
Yeah why not? I guess this is the new way of doing things. Thanks.

PR #17, opened 2015-07-07
Add ESLint config with all latest v0.24 rules

nkbt:


nkbt:
## ESLint run

eslint

Most of these warnings/errors can be very quickly fixed

Also I did not do it here, but it is very cool to have eslint-loader added to webpack, so you can see warnings while you develop in realtime. Will add PR after this one.

chenglou:
I feel like this is way too much for the scope of what we have right now. I have local eslint config set up, it prevents a few common mistakes, but adding 200+ lines of lint makes no one wants to go back changing it; and I doubt it'll catch much useful lints.

nkbt:
@chenglou it actually catches pretty much all most common mistakes. It is large because it has all rules in there with descriptions. It may be unnecessary when you work alone but as we found in our company it works pretty well if you have distributed team. Additional comments are also very helpful to figure out why something is failing without going to the webz

It is up to you of course. Feel free to close. I just highly recommend to have a code style in place to avoid mistakes made by other contributors. And yourself as well =)

nkbt:
It is also very good to have when you add CI, so PR will just fail if it has style issues in there

huang47:
I agreed with @chenglou, this is way too much and didn't see too much benefit from it. something like use `const` instead var because the value never changed is also opinionated and slow down the development in this early stage.

gaearon:
I'd just adopt a widely spread convention (e.g. airbnb javascript eslint rules).
Like @DenisIzmaylov did here: https://github.com/gaearon/redux/blob/4f620e6aa5d21a78c7e624725230a84645588c12/.eslintrc#L2

iclanzan:
I really think we need this even at this stage. It will save us a lot of headaches down the road. Adopting AirBnB’s rules as @gaearon suggested I think is a very good idea.

chenglou:
@nkbt would you mind doing that? What @gaearon said

nkbt:
@chenglou for sure, np

package.json
```diff
@@ -9,7 +9,11 @@
   "devDependencies": {
     "babel": "^5.6.14",
     "babel-core": "^5.2.17",
+    "babel-eslint": "^3.1.23",
```
nkbt: Had to add this due to https://github.com/airbnb/javascript/pull/415

nkbt:
@chenglou done, see updated code.

chenglou:
Great! Thanks =)

nkbt:
\o/

PR #13, opened 2015-07-06
Production-ready React module setup

nkbt:
1. I've added ESLint with pretty strict config. I updated config to match your code style. Fixed only some actual errors and minor style inconsistency
2. Added karma with one empty test. It also gives you a coverage
3. Split webpack config to production and development for easier support and extension in a future
4. Added ESLint loader, so it will throw warnings/errors while you develop things in dev mode
5. changed __DEV__ to process.env.NODE_ENV, since the same const is used in React itself to check if PropTypes checks should be done
6. Added bunch of scripts including prepublish, so you can publish module on npm now already built properly and without any extras (npmignore)

Should pretty much fix #11

PR #19, opened 2015-07-07
Testing base, ES6 coverage report

nkbt:
As part of #13.
Independent from the code itself. Only had to fix same syntax issue in Spring, since parser crashed on that one.

When lib files are moved, these configs should be updated accordingly, but for the moment everything works fine.


nkbt:
## Tests run
tests

## Coverage report

chenglou:
I'll hold this off since we don't need jasmine, karma and code coverage right now. License's a no-brainer though.

nkbt:
No problem, it's quite isolated, so can be added anytime. If you prefer
another stack, let me know. Mocha, testem, tape, etc.

I recommend to solve this sooner then later, so it will be easy to write
tests and encourage contributors to increase coverage.

iclanzan:
I would also love to see this solved asap. I personally prefer Testem + Tape, but anything will do.

chenglou:
Let's do mocha for now because it's the one I'm familiar with. We can check into the alternatives later.

(cc @dariocravero who had a huge PR with some tests)

gaearon:
In my experience Mocha is amazing, + Karma if you need DOM.

chenglou:
:+1 If you say so! Don't think there's enough DOM-related logic so we can keep karma for later. Mocha it is for now then!

nkbt:
Deal, will use Mocha and update PR asap.

chenglou:
@nkbt can you keep karma and coverage out for now?

nkbt:
@chenglou sure, no karma. That was just a rebase commit.

nkbt:
@chenglou why not coverage? It works pretty well. The only thing I needed to do is to rename Spring.jsx to Spring.js, so it appears in coverage report (spent decent amount of time trying to find workaround to get report for js and jsx, messing with isparta and istanbul, but found the best solution is just to rename it to js)

chenglou:
Meh. I guess we can keep coverage for now. When you're done, please squash and I'll merge

nkbt:
Squashed.

chenglou:
:+1: thanks!

nkbt:
Great!

PR #23, opened 2015-07-08
Remove hack

mirkodrummer:
@chenglou Since I saw your talk at React Europe(I was there) and I wanted to use Spring asap, this pull request is aimed to remove the hack, because I basically don't want my user slow down animations on fat keydowns :D

### DIFF
--- a/Spring.jsx
+++ b/Spring.jsx
@@ -386,7 +386,7 @@ export let TransitionSpring = React.createClass({
       let mergedVals = mergeDiffObj(
         currVals,
         endValue,
-        key => willLeave(key, endValue, currVals, currV),
+        key => willLeave(key, endValue, currVals, currV)
       );
 
       currVals = clone(currVals);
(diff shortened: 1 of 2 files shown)


Spring.jsx
```diff
@@ -386,7 +303,7 @@ export let TransitionSpring = React.createClass({
       let mergedVals = mergeDiffObj(
         currVals,
         endValue,
-        key => willLeave(key, endValue, currVals, currV),
+        key => willLeave(key, endValue, currVals, currV)
```
chenglou: I'll let this be but I might put it back another time for making arguments easier to move and around and diff. Babel takes this and removes it anyways for backward compat

chenglou:
Hah, can you believe some people are already using it? Wait till their users find out pressing on a key wreaks havoc in their app =P

Thanks for the PR. We'll make an actual dedicated functionality for this in the future.

Spring.jsx
mirkodrummer: I removed it because browserify report a build error :D

chenglou: Maybe it's because you didn't enable the comma transform in babel? It's es7 I think. Anyhow, the real library will ship the transpired version so we'll be fine

PR #25, opened 2015-07-09
Syntax fixes

nkbt:
Unused vars, dead code, redefined vars


chenglou:
Welp, guess it's time to start using `const` =). Thanks!

PR #28, opened 2015-07-09
adding null check for endValue because sometimes payloads have nulls

jlebensold:
When a dto has null values, the recursive calls on the payload cause the calculation to explode on a null check.


chenglou:
Thanks!

Issue #31, opened 2015-07-09
Switching between browser tabs


chtefi:
Hi,

Thanks for this awesome lib.

I stumbled upon something suspect when switching between browser tabs while an animation is on.
The following code simply renders an interpolated value (0 at first, then 300 after 1s) and use it in a  `transformX` as well.
Everything works fine.

Until you switch your browser tab, come back to it, switch, come back etc. The value keep changing/bouncing a lot at each switch (due to the low damping it seems? 5 in this example)
Is that expected?

![](http://i.gyazo.com/cf5489de99ba953fb22d5ea81efa8d4e.gif)

```javascript
export default class App extends React.Component {
	constructor() {
		super();
		this.state = { value: 0 };
		setTimeout(() => this.setState({ value: 300 }), 1000);
	}
	render() {
		return (
			<Spring endValue={{val: this.state.value, config: [ 100, 5 ] }}>
		  		{interpolated => <div style={{
              		 	  transform: `translateX(${interpolated.val}px)`,              
           	 	}}>{~~interpolated.val}</div>}
			</Spring>
		);
	}
}

```

PS: I'm running on Chrome 44 Windows.

KyleAMathews:
Browsers throttle requestAnimationFrame (what react-motion uses to power its animations) significantly when a tab is inactive ([this thread](http://greensock.com/forums/topic/9887-tween-paused-when-switch-to-other-tabs/) says ~2 frames/second).

I think the normal thing to do here is to pause the animation when you detect the user leaves the tab and restart it when they return.

KyleAMathews:
Though it'd be cool if react-motion could handle this somehow automatically.

nkbt:
It is fairly easy to check if tab is in foreground or background. I'll have a look today (after work) what could be done if no one fixes it before that

chenglou:
What's really the desired behavior here? Slow down or stay at the same speed?

nkbt:
I am pretty sure animation should freeze and resume when we are back.

>

chenglou:
Well maybe this is a bad analogy, but we don't do the same with sound right? And what's really "going away", does losing focus by switching to another app (while the screen is still partially visible) count? Because if so it seems that continuing the animation's the way to go? Not talking about technical details here.

nkbt:
Analogy with sound is not that straight. You cannot see animation when tab
is not visible. Background tab is an issue though. Visible **and**
inactive.

[...]

PR #39, opened 2015-07-10
Use semi-implicit Euler integration.

iclanzan:
More accurate and much more stable. <https://en.wikipedia.org/wiki/Semi-implicit_Euler_method>

### DIFF
--- a/stepper.js
+++ b/stepper.js
@@ -15,8 +15,8 @@ export default function stepper(frameRate, x, v, destX, k, b) {
   // let a = (Fspring + Fdamper) / mass;
   const a = Fspring + Fdamper;
 
-  const newX = x + v * frameRate;
   const newV = v + a * frameRate;
+  const newX = x + newV * frameRate;
 
   if (Math.abs(newV - v) < errorMargin && Math.abs(newX - x) < errorMargin) {
     return [destX, 0];


iclanzan:
Consider the old code `const newX = x + v * frameRate;`. Assuming no initial velocity `x` would remain unchanged after the first step.

chenglou:
Hah, physics =)
Curious though, how does this change make it "semi-implicit"? The wiki article's a bit hardcore

iclanzan:
I don’t know why it is named “semi-implicit” but the difference to the standard Euler is that you use either the current velocity to to compute the current position or the current position to compute the current velocity. I chose the first approach to avoid the position staying the same after the first step.

Here are the methods exemplified in the context of our `stepper.js` file:

```js
// Euler integration
const newX = x + v * frameRate;
const newV = v + a * frameRate;

// Semi-implicit Euler (1st variation)
const newV = v + a * frameRate;
const newX = x + newV* frameRate;

// Semi-implicit Euler (2nd variation)
const newX = x + v * frameRate;
// Here you would need to compute the acceleration `a` using `newX`.
const newV = v + a * frameRate;
```

iclanzan:
Also, the frameRate needs to be constant for best results with integration whereas right now the frameRate is all over the place because `requestAnimationFrame` does not fire at a constant interval.

For this reason it would make sense to decouple the stepper from rendering.
You might find this useful: <http://gafferongames.com/game-physics/fix-your-timestep/>.

razh:
Following up on a related article, thoughts on using RK4 for integration: http://gafferongames.com/game-physics/integration-basics/?

Here's a visualization of RK4:
rk4_diagram
(source: http://drewskillman.com/thesis.html)

Example implementation (correct me if I'm wrong):
```js
const xa = x;
const va = v;
const aa = -k * (xa - destX) - b * va;

const xb = x + va * dt / 2;
const vb = v + aa * dt / 2;
const ab = -k * (xb - destX) - b * vb;

const xc = x + vb * dt / 2;
const vc = v + ab * dt / 2;
const ac = -k * (xc - destX) - b * vc;

const xd = x + vc * dt;
const vd = v + ac * dt;
const ad = -k * (xd - destX) - b * vd;

const dxdt = (va + 2 * (vb + vc) + vd) / 6;
const dvdt = (aa + 2 * (ab + ac) + ad) / 6;

const newX = x + dxdt * dt;
const newV = v + dvdt * dt;

return [newX, newV];
```

Where `frameRate` is replaced with `dt`.


If we add a helper for calculating acceleration:
```js
const acceleration = (x, v) => -k * (x - destX) - b * v;

const xa = x;
const va = v;
const aa = acceleration(xa, va);

const xb = x + va * dt / 2;
const vb = v + aa * dt / 2;
const ab = acceleration(xb, vb);

const xc = x + vb * dt / 2;
const vc = v + ab * dt / 2;
const ac = acceleration(xc, vc);

const xd = x + vc * dt;
const vd = v + ac * dt;
const ad = acceleration(xd, vd);
```

iclanzan:
The RK4 is 4 times as expensive computationally and I don’t think that extra bit of accuracy is worth it. But I am not against it either.

razh:
Did some quick testing, regular Euler integration never accumulates more than a few pixels worth of errors more than RK4 at the timescale/distances of most UI animations.

chenglou:
Gosh this is interesting. I'll read through this soon. For reference, Rebound-js uses RK4: https://github.com/facebook/rebound-js/blob/fe8994c1caca9f0e7ff27336c5ecf0a4abd807dc/rebound.js#L601-L605

chenglou:
Related: #31 which is causes by the weird time delta we get from browser slowing down rAF
Merging this. @iclanzan I'd love to decouple stepper from the rendering, gladly accepting PR =)

PR #40, opened 2015-07-10
Big refactor.

dariocravero:
In short:

- Extracted Spring and TransitionSpring into their own components.
- Introduced a Motion component that shares logic amongst them.
- Brought in redux scripts and build process.
- Moved utils into utils folder and into separate files. Commented out
  the ones that aren't being used.
- Tried to comply to as many eslint rules as possible :)
- Moved to using classes in components.
- Added empty tests.

I can update the PR with the recent changes if you are happy with the direction this is taking @chenglou!

This should set the base for a more modular approach :), thoughts?


src/spring.js
```diff
@@ -0,0 +1,54 @@
+import { createEndValueType, FRAME_RATE, noSpeed, updateCurrV, updateCurrVals } from './utils';
+import Motion from './motion';
+import { PropTypes } from 'react';
+
+export default class Spring extends Motion {
```
nkbt: Welcome to the OOP world =(. Inheriting inherited inheritance

nkbt: I spent at least couple of minutes to find out where the heck is `this.raf`...

dariocravero: I'm ok with have both Spring and TransitionSpring be as they were instead of extending a new class. However, it's almost 40 lines of code that they share and, unless they are to be completely different, that's a great deal of overhead. Motion works more like an abstract class in a way. If that pattern is kept, perhaps the [invariant check](https://github.com/dariocravero/react-motion/commit/f68cca96e3322b3d7b55110eaf290e76d277af77#diff-c333f12887279f59e771971f7607eb37R9) can be replaced by something like `raf() { throw Error('implement it!') }`.

nkbt: Common code can be extracted to functions. If it cannot be extracted, then core working component can be wrapped into two main components. Not inherited.

Considering most of the functionality is related to pure calc, it should be very easy to extract common code to pure functions.

chenglou:
I only know babelrc loose mode is faster but not sure what the edge cases are? Probably doesn't apply to this repo right?

Also this is way too big and has way too many conflicts lol, is it possible to have a PR for the babelrc and LICENSE? I can merge those easily

nkbt:
@chenglou done #46 #47

dariocravero:
@chenglou sorry about the size of the PR, I thought about chunk-sizing it too but it was too much trouble as one thing was depending on the other; mainly splitting Spring.jsx into multiple files and putting a better build process that outputs all the files to `lib`. I guess the question is: are you ok with the way the refactor was going? Or are you planning in keeping everything as it is?

chenglou:
Refactored a bit and the codebase actually has tests now (lol). Will gradually refactor more. I think it's safe to close this now =).

dariocravero:
I still think that Spring has to be split into more files but yes, it's a good start :). Let me know if you need some help with this!

PR #42, opened 2015-07-11
Add repository to package.json

cesarandreu:
Adding this means that npmjs.com will include a link to the repo here:

npmjs.com react-motion page

Example:

npmjs.com browserify page


chenglou:
Thanks!

Issue #20, opened 2015-07-07
Define custom component tag name


arcthur:
Hi,

I use the animation in react on different structure, not all of 'div', such as 'ul' or 'ol', so I want to the coder could be defined the component tag name.

chenglou:
You don't have the restriction here. Use any tag you want:
```jsx
<Spring endValue={{val: 10}}>
  {interpolated => <ul>{interpolated.val}</ul>}
</Spring>
```

Hope that helps! =)

PR #43, opened 2015-07-11
added component prop to allow custom tag

souporserious:
In relation to this issue https://github.com/chenglou/react-motion/issues/20 and inspired from the functionality of ReactTransitionGroup. I think this is a very valuable prop since someone might want to animate an unordered list for instance and not have extra tags floating around.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 .DS_Store
 node_modules
 all.js.map
-lib
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,4 @@
 react-motion.sublime-*
 server.js
 demo*
+bower.json
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # React-Motion
 
 ```js
-<Spring endValue={{val: 10}}>
+<Spring component="div" endValue={{val: 10}}>
   {interpolated => <div>{interpolated.val}</div>}
 </Spring>
 ```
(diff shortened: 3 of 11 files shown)


chenglou:
(Continued from Slack discussion)

React-motion doesn't have the constraint of TransitionGroup of needing a direct, flat list without wrappers right inside the component. This:

```jsx
<Spring ...>{vals => <ul>...</ul>}</Spring>
```

is no more tedious than this:

```jsx
<Spring component="ul" ...>{vals => ...}</Spring>
```

Former's actually clearer and doesn't require a dedicated API.

I get that your use-case is to create wrapper components on top of react-motion, of the format:

```jsx
<FadeIn>
  <li>...</li>
</FadeIn>
```

Where you want the FadeIn to become a `ul`. But frankly I've moved away from this because it felt inelegant (bad argument though...). Maybe [HoC](https://gist.github.com/sebmarkbage/ef0bf1f338a7182b6775) can better help you here? Once you've written the Enhance it should be just `<FadeInUl>` or something.

I'm closing this for now, for the same reason as #20. I don't feel the increased API surface is worth it.

(Relevant: https://www.youtube.com/watch?v=4anAwXYqLG8, something I stand by)

nkbt:
@chenglou you can use something like 
```js
React.cloneElement(React.Children.only(this.props.children), additionalProps);
```

This does not break React reconciliation and works fine.

Issue #33, opened 2015-07-10
Put some badges


chenglou:
Npm/Bower/Slack

nkbt:
devDependencies, CI (circleci?), code coverage (coveralls?)

[![Dependency Status](https://david-dm.org/chenglou/react-motion.svg)](https://david-dm.org/chenglou/react-motion) [![devDependency Status](https://david-dm.org/chenglou/react-motion/dev-status.svg)](https://david-dm.org/chenglou/react-motion#info=devDependencies)

```md

[![Dependency Status](https://david-dm.org/chenglou/react-motion.svg)](https://david-dm.org/chenglou/react-motion)

[![devDependency Status](https://david-dm.org/chenglou/react-motion/dev-status.svg)](https://david-dm.org/chenglou/react-motion#info=devDependencies)
```

chenglou:
Npm bower and slack are fine. Or maybe just slack. Having like 5+ badges is kinda overloading it

nkbt:
=)

you can always put the m the the end of readme. Still nice to have some visual indication that health is OK.

srcspider:
@chenglou Maybe add a gitter link for anything that might require irc-like live chat (help with anything for example). It's easy to use and unlike irc people easily see what people were chatting about before they arrived and can also answer someone even after they leave and the person will just receive an email about it. It's also visible to guests, and you just need github to access it. Here's webpack's as an example: https://gitter.im/webpack/webpack/

Maybe this is just a setting, but Slack presented me with a big "sign up to do anything", and didn't feel like doing that since they presented to me a "hidden steps form" for sign up, but I assume it does something similar to gitter but is more of a private thing for core contributors only?

nkbt:
:+1: for gitter 
:-1: for slack (http://blog.freecodecamp.com/2015/06/so-yeah-we-tried-slack-and-we-deeply-regretted-it.html)

PS: Gitter has IRC gateway https://irc.gitter.im/ which works perfect.

chenglou:
Tbh I'm not much of a slack user and I still find it weird, but I chose slack because the Reactiflux channel already has thousands of users and slack seems more popular than gitter. Popularity sounds like a weak argument but probably not in this case.

Is there any gitter -> slack or irc -> slack integration?

srcspider:
@nkbt didn't know that, awesum

gaearon:
I've heard Slack can be integrated with IRC.

PR #48, opened 2015-07-12
Add badges

johnamiahford:
Is there a way to target the `react-motion` channel specifically for the slack badge?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
 # React-Motion
 
-[![npm package](https://img.shields.io/npm/v/react-motion.svg?style=flat-square)](https://www.npmjs.org/package/react-motion)
-![Bower version](https://img.shields.io/bower/v/react-motion.svg?style=flat-square)
+[![npm version](https://badge.fury.io/js/react-motion.svg)](https://www.npmjs.com/package/react-motion)
 [![react-motion channel on slack](https://img.shields.io/badge/slack-react--motion%40reactiflux-61DAAA.svg?style=flat)](https://reactiflux.slack.com/messages/react-motion/)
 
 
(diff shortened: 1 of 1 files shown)


nkbt:
I would think twice using slack for open source https://github.com/chenglou/react-motion/issues/33#issuecomment-120673921

chenglou:
Wait I don't get how this diffed... The slack badge is already there. Weird?

nkbt:
@johnamiahford hasn't rebased his branch

chenglou:
Heh I thought that still wouldnt pass github's merge check. @johnamiahford can you rebase and add just the npm and bower badges please? Also, the existing slack badge should directly go to the channel already

nkbt:
I can't join slack since no one invited me to reactiflux. Is it how it works?

chenglou:
You don't need that (I know right, misleading...), just join with your email here: http://www.reactiflux.com

nkbt:
Thanks. I'm too used to gitter

johnamiahford:
Changed to shields.io badges to be consistent.

chenglou:
Argh those orange badges scream "don't come near". This is almost literally bike shedding but, is there a way to have the green/blue ones?

Also, use flat instead of flat-square to be consistent with the Slack badge please

johnamiahford:
Ha. I'll see if I can pretty these up for you.

johnamiahford:
I've removed the bower package as it was returning not found. And went with the badge fury npm package as it gives us that pretty green we all love.

chenglou:
Hmm that's weird, I'll take what's here then. Thanks!

PR #53, opened 2015-07-12
Lint fixes / Demo0

nkbt:
One-by one syntax/lint fixes...

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,6 +8,10 @@
   ],
   "rules": {
     "comma-dangle": [2, "always-multiline"],
+    "no-eq-null": 0,
+    "prefer-const": 2,
+    "block-scoped-var": 0, // until fixed https://github.com/eslint/eslint/issues/2253
+
 
     "react/jsx-no-undef": 2,
     "react/jsx-quotes": [2, "double", "avoid-escape"],
--- a/Spring.js
+++ b/Spring.js
@@ -1,5 +1,5 @@
 import React, {PropTypes} from 'react';
-import {range, mapTree, clone} from './utils';
+import {mapTree, clone} from './utils';
 import stepper from './stepper';
 
 // ---------
@@ -39,7 +39,7 @@ function mergeDiff(collA, collB, onRemove, accum) {
 }
 
 function mergeDiffObj(a, b, onRemove) {
-  const keys = mergeDiff(Object.keys(a), Object.keys(b), a => !onRemove(a), []);
+  const keys = mergeDiff(Object.keys(a), Object.keys(b), _a => !onRemove(_a), []);
   const ret = {};
   keys.forEach(key => {
     if (b.hasOwnProperty(key)) {
@@ -140,7 +140,7 @@ function noSpeed(coll) {
 }
 
 
-export default React.createClass({
+const Spring = React.createClass({
   propTypes: {
     endValue: PropTypes.oneOfType([
       PropTypes.func,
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
   "scripts": {
     "start": "NODE_ENV=development node server.js",
     "build": "NODE_ENV=production webpack -p",
-    "lint": "eslint .",
+    "lint": "eslint --ext .js,.jsx .",
```
nkbt: Completely forgot we have jsx here...

chenglou:
you can just fix all the demos at once I think

chenglou:
Also the triple equal rule is too strong. I use `a == null` to captured undefined too and that's the only place I use it.

nkbt:
Hm. Double equal is quite unsafe. It is better to be more explicit. If you want to capture undefined AND null, then it should be actully written like that =).

Will fix all demos, ok

chenglou:
Kinda used to it and Facebook internally allows it just fine. Don't worry, won't abuse it here.

nkbt:
Okay. Will update eslint config then.

.eslintrc
```diff
@@ -8,6 +8,10 @@
   ],
   "rules": {
     "comma-dangle": [2, "always-multiline"],
+    "no-eq-null": 0,
+    "prefer-const": 2,
```
nkbt: need to make pr to airbnb =)

Spring.js
```diff
@@ -140,7 +140,7 @@ function noSpeed(coll) {
 }
 
 
-export default React.createClass({
+const Spring = React.createClass({
```
nkbt: We should have var so react can find out the component name.

utils.js
```diff
@@ -1,5 +1,3 @@
-'use strict';
-
 // damn it JS
 export function clone(coll) {
   return JSON.parse(JSON.stringify(coll));
```
nkbt: Better not to do this thing. Horrible performance.

chenglou: Oh I'm very aware of that =) It probably prevented all sort of bugs during iteration for me. I have a todo to remove those once I'm done.

chenglou: We also allocate like crazy in Spring. All these will be refactored once things settle down, so don't worry about it for now.

chenglou:
(Is this done?)

nkbt:
Yep. Walked through all modules and lint report is nice and clean now.

utils.js
nkbt: Yeah, didn't touch it now of course.

demo3/Demo.jsx
```diff
@@ -148,12 +149,12 @@ let Demo = React.createClass({
                           type="checkbox"
                           onChange={this.handleDone.bind(null, date)}
                           checked={isDone}
-                        />
+                          />
```
chenglou: Is this a lint rule? Old one's fine?

nkbt: Nah... I think it is my autoformatter. Do you think it should not be all aligned?

chenglou: Always been coding in the old style (fb's internal style)

nkbt: Btw I usually keep closing bracket on the last line, so it does not look awkward.
Like:

```js
                        <button
                          className="destroy"
                          onClick={this.handleDestroy.bind(null, date)} />
```

chenglou: Same, but got used to the internal convention

nkbt: ok. reverted that change.

chenglou:
Merging! Btw the style nits really have nothing to do with whatever "internal rules". Just got used to them personally and didn't want to juggle between them =)

PR #56, opened 2015-07-12
Fix value for viewport

cody:


nkbt:
:+1:

chenglou:
So there were no such thing as just `scale`?
(Hey cody, I remember your profile picture =))

cody:
@chenglou 
Safari complains that
> Viewport argument key "scale" not recognized and ignored.

and Chrome says:
> The key "scale" is not recognized and ignored.

The possible values for viewport are documented at MDN:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta

chenglou:
:+1:

PR #64, opened 2015-07-14
Added the ability to change the wrapper component from the default "div"

bringking:
This PR adds the ability to specify the wrapper component on Spring and TransitionSpring rather than let it default to a DIV. This is very similar to the way the current ReactTransitionGroup works. This will allow ReactMotion to work semantically in more places, as well as work inside Tables without causing invariant violations. I updated the readme to reflect the changes and added unit tests into Spring-test.js. Let me know if this looks okay! Thanks for the awesome work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,9 @@ lib
 
 .DS_Store
 all.js.map
+
+
+# Intellij
+.idea/
+*.iml
+*.iws
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ getEndValues: function(currentPositions) {
 },
 
 ```
-Spring also exposes an optional "component" prop similar to [ReactTransitionGroup](https://facebook.github.io/react/docs/animation.html#rendering-a-different-component) that allows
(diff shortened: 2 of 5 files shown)


chenglou:
Hey! Check this commit: https://github.com/chenglou/react-motion/commit/8fc29c26c3915150a4638c520516b5f9e2817921

Neither Springs have the wrapper now, so you get what you want without a dedicated API =).
Closing unless I'm missing something. Thanks though!

bringking:
Hmm maybe I am confused, but returning  ```React.Children.only(this.props.children(currVals))``` still results in a <div> wrapper in the DOM. 
screen shot 2015-07-14 at 10 17 42 am

This PR allows you to change that to a SPAN or LI or whatever element you want.

chenglou:
Oh, those are the old compiled files I didn't bother to build again from the new commits, sorry about that! I didn't want to check in huge files each time I make a change. Since the end result looks the same visually I will only be updating them once in a while.

bringking:
Ah I see, the demo's are running off old artifacts, got it.

PR #71, opened 2015-07-16
Current positions is undefined at first render

peterssonjesper:
A bit nitty gritty, but it confused me when trying react-motion for the first time. I tried to do a similar thing as in this demo, but I used === instead (which doesn't eval as true for undefined and null, but == does).

### DIFF
--- a/demo1/Demo.jsx
+++ b/demo1/Demo.jsx
@@ -17,7 +17,7 @@ const Demo = React.createClass({
 
   getValues(currentPositions) {
     // currentPositions of `undefined` means it's the first render for Spring
-    if (currentPositions === undefined) {
+    if (currentPositions == undefined) {
       return {val: range(6).map(() => [0, 0])};
     }
     const endValue = currentPositions.val.map((_, i) => {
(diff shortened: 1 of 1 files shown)


nkbt:
now at first render you check if there are any values and set your initial value if you want one.

nkbt:
kind of
```js
  endValue(...args) {
    if (!args.length) {
      return {val: 0};
    }
    return {val: this.state.get('shouldClose') ? 0 : 1};
  },

// ...
      <Spring endValue={this.endValue}>
```

chenglou:
Humm, `==` is used (and always used) for `null` and `undefined` check at once, and nowhere else. I'd like to keep it this way. Can you just change the comment to `undefined`?

chenglou:
This does bring an issue that I've been thinking about. People keep asking how to do animation at mounting. This is the way to do it but it's not obvious enough. Maybe I _will_ release the `initialValue` prop after all.

nkbt:
initial value is the great idea. I was thinking about adding it right 1 hour ago =)

peterssonjesper:
@nkbt Yep, I realized that. Just confusing that the demo explicitly said that the argument is supposed to be null.

@chenglou I agree on the initialValue prop, that would probably be easier to understand.

chenglou:
Oh sorry, didn't see you updated the PR. Thanks!

PR #74, opened 2015-07-17
Expose dev server to the network.

iclanzan:
You know, so we can easily test on other devices.


gaearon:
Note: this doesn't work on Windows IIRC.

iclanzan:
Yes, you are right.

PR #78, opened 2015-07-18
Replace filterRight with a custom filter that does not mutate.

iclanzan:
As @chenglou pointed out, `filterRight` was `O(n^2)` because of that call to `splice`, so the regular old `Array.prototype.filter` is in fact faster.

The performance gains I was seeing when I switched from `filter` to `filterRight` were caused by not inlining the predicate.

Since not inlining the predicate is a requirement, I can’t think of a way to use the built in filter and pass the alpha to the predicate. I could pass it as `this` to the predicate but that has the overhead of wrapping the number in a `Number` object drastically affecting performance.

This is the best thing I could come up with. Implement a `filter` function that passes the third argument as the first argument to the predicate. As a bonus it is faster than `Array.prototype.filter`.

I am happy not to have to mutate the array anymore.


test/filter-test.js
```diff
@@ -0,0 +1,24 @@
+import expect from 'expect';
+import filter from '../src/filter';
+
+describe('filterRight', () => {
```
chenglou: Just `filter` now

chenglou:
Great. Pulling this in in favor of #81, which is much more performant but less readable and bug-prone.

iclanzan:
Good choice :+1:

PR #81, opened 2015-07-18
Remove filterRight.

iclanzan:


src/animationLoop.js
```diff
@@ -74,16 +68,31 @@ const prototype = {
...
-      renderSubscriber,
-      1 + animationLoop.accumulatedTime / timeStep
-    );
+    const alpha = 1 + animationLoop.accumulatedTime / timeStep;
+    let index = 0;
+    let nextState;
+    let subscriber;
+
+    while (index < state.length) {
+      subscriber = state[index];
```
chenglou: If you inline the `let` here babel will transform it into a regular `var`. Probably _slightly_ more readable.

chenglou:
As discussed, this is indeed less readable. But could allocate a lot less if you do that opt where you check if all subscribers are active right?

iclanzan:
Absolutely. I think it’s a nice optimization.

chenglou:
So are we gonna do the optimization or leave it for later?

iclanzan:
What do you mean? The optimization is right there. Line 84 only allocates a new array if a subscriber is not active.

iclanzan:
Looking at it I realize it’s flawed. It would miss adding active subscribers that are positioned before the first inactive one.

chenglou:
Yeah I think ironically, I got confused but this code and thought you were doing something else.

chenglou:
So, do you still think it's worth doing this or use the `filterRight`?

iclanzan:
Ok, fixed it.

iclanzan:
We should definitely **not** use `filterRight` since it’s `O(n^2)`.

chenglou:
Sorry meant to say #78

iclanzan:
It’s up to you. That one is more readable and easier to test. This one is more optimized.

src/animationLoop.js
```diff
@@ -32,6 +25,13 @@ const prototype = {
         subscriber.prevValue = value;
         subscriber.value = subscriber.step(timeStep, value);
       }
+
+      // Can’t use else here because subscriber.step() might cause
+      // subscriber to become inactive.
+      if (!subscriber.active) {
+        // Mark as dirty to perform filtering in the render step.
+        animationLoop.dirty = true;
```
nkbt: Can't see where `dirty` is reset. Is it a bug or feature?

src/animationLoop.js
```diff
@@ -74,16 +75,30 @@ const prototype = {
...
-      1 + animationLoop.accumulatedTime / timeStep
-    );
+    const alpha = 1 + animationLoop.accumulatedTime / timeStep;
+    let index = 0;
+    let nextState;
+
+    // If there are any inactive subscribers (loop is dirty)
+    // we need to filter state.
+    if (animationLoop.dirty) {
+      nextState = animationLoop.state = [];
```
nkbt: I would avoid double assignments.

src/animationLoop.js
```diff
@@ -74,16 +75,30 @@ const prototype = {
...
+    }
+
+    while (index < state.length) {
```
nkbt: thous should perfectly work with old good `for` loop instead.

iclanzan: True, I just have a preference for `while`.

src/animationLoop.js
iclanzan: Are you saying you would rather have the following?
```
nextState = [];
animationLoop.state = nextState;
```

src/animationLoop.js
nkbt: Yeah, you've implemented `for` loop with incremented guard with `while` :).

src/animationLoop.js
iclanzan: Bug. Good catch!

chenglou:
Keeping this open as a task for when we get enough test coverage to replace #78. We'll get there eventually =)

src/animationLoop.js
nkbt:
```js
nextState = [];
animationLoop.state = [];
```
If you want to actually keep mutable reference, then yes
```js
nextState = [];
animationLoop.state = nextState;
```

chenglou:
Diverged, closing but will use as reference.

PR #94, opened 2015-07-22
Add Karma

nkbt:


test/filter-test.js
```diff
@@ -9,8 +8,8 @@ describe('filter', () => {
 
     expect(result).toContain(1);
     expect(result).toContain(3);
-    expect(result).toNotContain(2);
-    expect(result).toNotContain(4);
+    expect(result).not.toContain(2);
```
nkbt: !This was the only difference after replacing expect with jasmine

karma.conf.js
```diff
@@ -0,0 +1,69 @@
...
+        {type: 'text-summary', file: 'text-summary.txt'},
+      ],
+    },
+    captureTimeout: 90000,
+    browserNoActivityTimeout: 60000,
+    port: 9876,
+    colors: true,
+    logLevel: config.LOG_INFO,
+    autoWatch: false,
+    browsers: ['PhantomJS'],
```
nkbt: Only phantom.
If you wish to debug tests in browsers, simply run `npm run test:dev` and point your browser (any) to http://localhost:9876 (karma port).

chenglou:
Ehh, does jasmine support good async testing like mocha's `done` passed into the callback? Because we're probably gonna need it.

nkbt:
Sure, it does, but it is always better to use spies and mocked time instead.

```js
  it("should support async execution of test preparation and expectations", function(done) {
    value++;
    expect(value).toBeGreaterThan(0);
    done();
  });
```

chenglou:
Cool, that's jasmine 2.0 right. It's been a while but I recall jasmine being on < 2.0 for quite a while.
Will mocked time work correctly with the new animation loop? Say, I want fine-grained simulation for when the next raf is. And I assume this works in phantom?

nkbt:
Mocking works pretty well. Phantom, yes. I would not submit a PR that does
not work lol :)

package.json
```diff
@@ -28,9 +34,10 @@
     "build-demos": "NODE_ENV=production webpack -p",
     "lint": "eslint --ext .js,.jsx .",
     "prerelease": "babel src --out-dir lib && NODE_ENV=production webpack --config webpack.prod.config.js",
-    "test": "mocha --compilers js:babel/register --recursive test",
-    "test:dev": "npm test -- --watch",
-    "test:cov": "`npm bin`/babel-node `npm bin`/isparta cover --report text --report html `npm bin`/_mocha -- --recursive src"
+    "karma": "karma start ./karma.conf.js",
+    "test": "npm run karma --  --single-run --reporters nyan",
+    "test:dev": "npm run karma -- --no-single-run --auto-watch --reporters nyan",
```
chenglou: I asked for a minimal PR. Is nyan cat minimal?!

jkjk =)

chenglou:
:+1:

PR #96, opened 2015-07-23
corrects imports of lodash.isplainobject module

balanceiskey:
Hi there!

When using react-motion in case-sensitive environments (Linux/Ubuntu), we run into an issue where node, webpack and possibly others can't properly require in the `lodash.isplainobject` module due to current spelling. Note, we've worked around this issue for now by simply aliasing inside of webpack but ideally I think we should use the true name of the module.

### DIFF
--- a/src/Spring.js
+++ b/src/Spring.js
@@ -1,6 +1,6 @@
 import React, {PropTypes} from 'react';
 import mapTree from './mapTree';
-import isPlainObject from 'lodash.isPlainObject';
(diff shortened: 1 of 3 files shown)


chenglou:
Well you won't need your workaround anymore =)

Sorry about that. Had the same issue with `require('React')` too. Somehow forgot about this.

balanceiskey:
Yay! Thanks for the quick merge and for this awesome library!

PR #97, opened 2015-07-23
Iterate over items added during filtering.

### LINKED ISSUE #95: Nesting two springs doesn't work
cc @iclanzan because I not sure whether that's a flaw in the current implementation of animation loop of the problem of the (which would be a bigger problem).

Edit: no, I think it's just a flaw in the current implementation.

iclanzan:
This should fix #95.


chenglou:
@bsansouci

chenglou:
:+1:

PR #99, opened 2015-07-24
Fix inject-loader

nkbt:


test/Spring-test.js
```diff
@@ -1,22 +1,23 @@
...
+
+const injector = require('inject!../src/Spring');
 
 const FRAME_RATE = 1 / 60;
 
 describe('updateCurrValue', () => {
+  let updateCurrValue;
+
+  beforeEach(() => {
+    updateCurrValue = injector({
```
nkbt: I highly recommend to create instances for each test, so spies clear their state. Usually you put some spy as dependency and then can find out it is called as expected with correct data. If you don't create them for each test, spy calls counter will increment every time.

chenglou:
:+1:

PR #102, opened 2015-07-26
Refactor animationLoop. Extract logic from both Spring.

bsansouci:
This is an attempt at making the `animationLoop` simpler to reason about.

`animationStep` is now outside either `Spring`s as a pure function with
the following signature: `(timestep, state) => newState`
- `timestep` is the timestep given by the `animationLoop`
- `state` is the cached state inside the `animationLoop`.(which won't
  be in sync if you call `setState` while the loop is running)

`subscribeAnimation` is now called `startAnimation`, and allows being
called multiple times in a row. This is so we can stop checking if the
animation was already started in the Spring. We still need to keep
`this.stopAnimation` for the sole purpose of cancelling the animation
when the Spring gets unmounted.

You can create an new instance of the `animationLoop` by calling
`configAnimation`. This takes a configuration object, making it easy
to test (by giving a different `raf`/`now`).

`startAnimation` still takes the state, a step function and a render
function.

Fix bug with `TransitionSpring`. Because the TS sets `destVal` as
`enterVals` for the first tick, we render that value twice. We render it once,
`currValue` is `enterVal` and `destVal` is `enterVal`. The next tick is faster
than 16ms, so `accumulatedTime` (which would be about -16ms from the previous
tick) is negative (-16ms + any number less than 16ms < 0). So we just render
part ways towards the `nextState`, but that's `enterVal` still. We render say
75% between `currValue` (=== `enterVal`) and `destValue` (=== `enterVal`). So we
render the same value a second time.


chenglou:
Feels a bit too big as a single PR. Hopefully you tested all the edge cases. Thanks!

bsansouci:
I tried to make it as small as possible, but I did change quite a bit of code. I've been testing all the edge cases that I could come up with. More concrete tests will help in the future.

PR #103, opened 2015-07-26
Demo5 - Fix mousemove outside of region. Fix for mobile.

bsansouci:


demos/demo5/Demo.jsx
```diff
@@ -18,6 +18,17 @@ const Demo = React.createClass({
     };
   },
 
+  componentWillMount() {
+    document.addEventListener('mousemove', this.handleMouseMove);
```
chenglou: You can try `mouseLeave` on the root component instead.

bsansouci: What would be the expected behavior of `mouseLeave`? I can just scale the main div to the screen size

chenglou: yeah

Issue #104, opened 2015-07-27
TransitionSpring stops even though `willLeave` never returns `null`


chenglou:
@bsansouci

PR #105, opened 2015-07-28
Make `willLeave` stop when `currVelocity` is 0 and `currValue` reached dest

bsansouci:
This makes `willLeave` simpler and have a better default behavior. The way to
tell the `TransitionSpring` before this was to add an if statement at the top of
`willLeave` to check if `currValue` has reached the destination and that the
`currVelocity` is 0 and then returning null if yes. We now do that check
internally and do the unmounting when `currValue` is equal to the return of
`endValue` and `currVelocity` is 0.

This "fixes" #104 by just deprecating the API.

### DIFF
--- a/demos/demo3/Demo.jsx
+++ b/demos/demo3/Demo.jsx
@@ -106,13 +106,11 @@ const Demo = React.createClass({
 
   // TODO: change naming
   willLeave(date, valueThatJustLeft) {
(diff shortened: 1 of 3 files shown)


chenglou:
We'll keep `null` around internally for perf, e.g. default `willLeave` returns null to signal unmount still

PR #107, opened 2015-07-29
react-native 'compatibility'

threepointone:
tests pass, demos work, doesn't break regular web webpack builds.

I just want my name in this repo

### DIFF
--- a/src/Spring.js
+++ b/src/Spring.js
@@ -1,4 +1,14 @@
-import React, {PropTypes} from 'react';
+
+let React;
+try {
+  React = require('react-native');
+} catch(e) {
+  React = require('react');
+}
+
+let {PropTypes} = React;
+
+
 import mapTree from './mapTree';
 import noVelocity from './noVelocity';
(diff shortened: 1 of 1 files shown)


chenglou:
did you check that the `build/` version also works? Script tag require with global ReactMotion

threepointone:
yes, just tested it, works fine. for posterity's sake, this was my file 

```html
<html>
  <head></head>
  <body>
    <script src="https://fb.me/react-0.13.3.js"></script>
    <script src='build/react-motion.js'></script>
    <script>
      console.log(ReactMotion);
    </script>
  </body>
</html>

```

chenglou:
We should probably automate that check. Thanks!

chenglou:
Humm webpack complains though. Is it possible to suppress the warnings? Like, extern react-native?

gaearon:
@chenglou You can have separate export for React Native and turn components into factories. For example https://github.com/gaearon/react-redux/blob/master/src/native.js.

PR #108, opened 2015-07-29
Inline animationloop

bsansouci:
This might be temporary as it makes the `animationLoop` impossible to
test on its own. 

The goal is to reduce the complexity of the system. It is a step towards 
implementing the prop `currVelocity` and maybe `currValue`. It also 
enables hot-reloading (as before closures where preventing hot-reloading 
to work correctly).

We're basically putting the animationLoop inside the `Spring` and
`TransitionSpring`, while keeping the separation between the
`animationLoop`, the `animationStep` and the render. We attached all the
properties that are needed for the loop onto `this`. We can't rely on
`this.setState` to set the state immediately so we save the `currValue`
into `currFrameValue` and `nextFrameValue` (the same way before we
had `animRunning[i].nextState` and `animRunning[i].prevState`).

Need to find a way to make `timeStep`, `timeScale` and `maxSteps`
configurable.


chenglou:
Done= 0>

PR #111, opened 2015-08-02
expose animationStep

threepointone:
expose animationStep (but somehow privately?) so you can port this to other frameworks (deku comes to mind)

### DIFF
--- a/src/components.js
+++ b/src/components.js
@@ -9,7 +9,7 @@ import {interpolateValue, updateCurrValue, updateCurrVelocity} from './updateTre
 
 const startAnimation = configAnimation();
 
-function animationStep(shouldMerge, stopAnimation, getProps, timestep, state) {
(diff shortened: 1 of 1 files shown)


chenglou:
This clashes with #108 which I might prioritize. `animationLoop` retains some values that we might need access to in the future. I'm sure we can still pull out big parts of this in the future, but right now I'm more concerned with enabling the rest of the APIs we're thinking about (e.g. exposing `currVelocity`) before doing the modularization.

chenglou:
What animationStep?

=)

Thanks a lot for the PR btw. The current status is that I've rolled back the animation loop into the respective components, as I've found out that there's actually not that much to be abstracted from them. We've previously done this but this has just caused a few more layers of indirections and perf cost without the benefit of abstracting away as much code as we've hoped.

But we should try this again some day!

PR #112, opened 2015-08-02
Passing scalars to `endValue`

bsansouci:
This will allow something like:
```js
<Spring defaultValue={0} endValue={100}>
{val => <div >{val}</div>}
</Spring>
```

We're overloading `endValue` and stepping away from the consistency we had before with `val` and `config`. But this use case seems common enough, and explicit enough, that react-motion could handle it. There is very little confusion as to whether the `Spring` will or not interpolate the _only_ thing it's given.

### DIFF
--- a/demos/demo0/Demo.jsx
+++ b/demos/demo0/Demo.jsx
@@ -24,8 +24,8 @@ const Demo = React.createClass({
           Toggle
         </button>
 
-        <Spring endValue={{val: this.state.open ? 400 : 0}}>
-          {({val}) =>
+        <Spring endValue={this.state.open ? 400 : 0}>
+          {val =>
             // children is a callback which should accept the current value of
             // `endValue`
             <div className="demo0">
--- a/src/updateTree.js
+++ b/src/updateTree.js
@@ -40,8 +40,8 @@ export function interpolateValue(alpha, nextValue, prevValue) {
   return nextValue;
 }
 
-// TODO: refactor common logic with updateCurrVelocity
-export function updateCurrValue(frameRate, currValue, currVelocity, endValue, k, b) {
+// TODO: refactor common logic with _updateCurrVelocity
+function _updateCurrValue(frameRate, currValue, currVelocity, endValue, k, b) {
   if (endValue == null) {
     return null;
   }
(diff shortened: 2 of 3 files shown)


demos/demo0/Demo.jsx
```diff
@@ -24,8 +24,8 @@ const Demo = React.createClass({
           Toggle
         </button>
 
-        <Spring endValue={{val: this.state.open ? 400 : 0}}>
-          {({val}) =>
+        <Spring endValue={this.state.open ? 400 : 0}>
+          {(val) =>
```
chenglou: strip parens here please

test/Spring-test.js
```diff
@@ -111,9 +111,9 @@ describe('Spring', () => {
           <Spring defaultValue={0} endValue={10}>
             {val => {
               count.push(val);
+              // count.length being at least 3 we have at least one
+              // intermediary value
               if (count.length > 2 && count[count.length - 1] === 10) {
```
chenglou: Check some intermediate values please. L70, 71 show such thing

src/updateTree.js
```diff
@@ -66,18 +66,29 @@ export function updateCurrValue(frameRate, currValue, currVelocity, endValue, k,
...
       return ret;
     }, {});
   }
   return endValue;
 }
 
-export function updateCurrVelocity(frameRate, currValue, currVelocity, endValue, k, b) {
+export function updateCurrValue(frameRate, currValue, currVelocity, endValue, k, b) {
+  if (typeof endValue === 'number') {
+    const _k = k || presets.noWobble[0];
```
chenglou: This logic is wrong. If either k or b is null we should return the default. We don't fill a value randomly if only one of them is absent.

bsansouci: You mean if they're _both_ null, then we do what I did here, otherwise we return `endValue`?

chenglou: Either: https://github.com/chenglou/react-motion/blob/e2f67203d6f751908f92c93532c779fa8c0ccd15/src/updateTree.js#L49

bsansouci: They're both going to be null if `endValue` is a scalar though.

chenglou: yeah but let's not rely on that.

chenglou:
Awesome.

PR #115, opened 2015-08-03
Add mockRaf (combining mocks for `raf` and `performance-now`)

bsansouci:
The title says it all.

I also adapted the tests to use that mock `raf`. The main advantages of using a mock `raf` is that the tests become synchronous, making them much easier to reason about. They also make the whole system deterministic. This means that when we call `mockRaf.step()` 10 times with the same Spring (and same constants),  `currValue` will always take the same values in the same order.

There are two main functions that can be used:
- `mockRaf.step(ms = 16)` - will step "time" by one frame and by `ms` amount of milliseconds. This will call all `raf`s that were registered in the last frame.
- `mockRaf.manySteps(numberOfSteps = 1)` - will call `mockRaf.step` `numberOfSteps` times

The two other functions are the mock functions. They can be used like follow:
```js
const injector = require('inject!../src/components');
const injectorAnimationLoop = require('inject!../src/animationLoop');
const {Spring, TransitionSpring} = injector({
  './animationLoop': injectorAnimationLoop({
    raf: mockRaf.getRaf(), // Will return a raf function that will just queue the callback
    'performance-now': mockRaf.getNow(), // Will return a now() function
  }),
})(React);
```

### DIFF
--- a/test/Spring-test.js
+++ b/test/Spring-test.js
@@ -1,27 +1,28 @@
-import React, {addons} from 'react/addons';
+import React, {addons, PropTypes} from 'react/addons';
+
+import createMockRaf from './createMockRaf';
+
 const TestUtils = addons.TestUtils;
 
-const injector = require('inject!../src/Spring');
-// const injectorAnimationLoop = require('inject!../src/animationLoop');
+const injector = require('inject!../src/components');
+const injectorAnimationLoop = require('inject!../src/animationLoop');
 
 describe('Spring', () => {
   let Spring;
   let TransitionSpring;
+  let mockRaf;
+
   beforeEach(() => {
-    Spring = injector({
-      // './animationLoop': injectorAnimationLoop({
-      //   raf: cb => {
-      //     cb();
-      //   },
-      // }),
-    }).Spring;
-    TransitionSpring = injector({
-      // './animationLoop': injectorAnimationLoop({
-      //   raf: cb => {
-      //     cb();
-      //   },
-      // }),
-    }).TransitionSpring;
+    mockRaf = createMockRaf();
+    const bothSprings = injector({
+      './animationLoop': injectorAnimationLoop({
+        raf: mockRaf.raf,
+        'performance-now': mockRaf.now,
+      }),
+    })(React);
+
+    Spring = bothSprings.Spring;
+    TransitionSpring = bothSprings.TransitionSpring;
   });
 
   it('should allow returning null from children function', () => {
@@ -32,9 +33,12 @@ describe('Spring', () => {
       },
     });
     TestUtils.renderIntoDocument(<App />);
+  });
 
+  it('should allow returning null from children function TransitionSpring', () => {
     const App2 = React.createClass({
       render() {
+        // shouldn't throw here
         return (
           <TransitionSpring endValue={{a: {}}}>
             {() => null}
@@ -58,19 +62,14 @@ describe('Spring', () => {
     TestUtils.renderIntoDocument(<App />);
   });
 
-  it('should allow a defaultValue', done => {
+  it('should allow a defaultValue', () => {
     let count = [];
     const App = React.createClass({
       render() {
         return (
-          <Spring defaultValue={{val: 1}} endValue={{val: 10}}>
+          <Spring defaultValue={{val: 0}} endValue={{val: 10}}>
             {({val}) => {
               count.push(val);
-              if (count.length === 2) {
-                expect(count[0]).toBe(1);
-                expect(count[1]).toBeGreaterThan(1);
-                setTimeout(done, 0);
-              }
               return null;
             }}
           </Spring>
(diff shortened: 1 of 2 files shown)


test/Spring-test.js
```diff
@@ -58,19 +60,14 @@ describe('Spring', () => {
...
 
-  it('should allow a defaultValue', done => {
+  it('should allow a defaultValue', () => {
     let count = [];
     const App = React.createClass({
       render() {
         return (
-          <Spring defaultValue={{val: 1}} endValue={{val: 10}}>
+          <Spring defaultValue={{val: 0}} endValue={{val: 10}}>
             {({val}) => {
```
nkbt: `({val}) => count.push(val)`?

test/Spring-test.js
```diff
@@ -167,26 +258,64 @@ describe('Spring', () => {
       },
     });
     TestUtils.renderIntoDocument(<App />);
+
+    // Initial render
+    expect(count).toEqual([0]);
+    mockRaf.manySteps(8);
+    expect(count).toEqual([
+      0,
```
nkbt: I'd stick to one list and put it to the top of the file as const (it is always the same with 8 steps and from 0 to 10)

test/mockRaf.js
```diff
@@ -0,0 +1,35 @@
+export default {
```
nkbt: Return factory and create object with empty values `beforeEach`, so you will avoid potential value clashes and your tests will be completely independent.

test/Spring-test.js
chenglou: One list doesn't cover much though? I guess it doesn't matter here since if you get one list right you'd get others right

nkbt: Exactly. If logic is correct, then all results will be the same and you can avoid repeating this fair bit of code all over the tests

test/mockRaf.js
bsansouci: Whoops good call, I wanted to do that but just forgot.

test/Spring-test.js
bsansouci: To keep track of all the values given to the child-as-function of the `Spring`. Or is that not your question?

test/mockRaf.js
nkbt: This one is the only concern. All other comments - just nit fixes.

test/Spring-test.js
bsansouci: I thought about that but then preferred being very explicit. I can put it at the top if you guys prefer.

test/mockRaf.js
```diff
@@ -0,0 +1,35 @@
...
+  allCallbacks: [],
+  prevTime: 0,
+
+  getNow() {
+    return () => {
+      if (this.prevTime === 0) {
+        this.prevTime = Date.now();
```
nkbt: I would avoid Date.now() by all means in tests. Stick to some static value instead. Working with dates we usually use something like `1000` as timestamp. So it is not zero, and not too long to type.

test/Spring-test.js
nkbt: I don't have a strong opinion on this matter. Personally prefer to put values in each test. DRY is not for tests. Maybe you can reduce steps to 4, for example, so you will have less numbers to repeat. 

Oh and another concern - precision. It could vary. Maybe you should round values to, to say, 4-5 digits?

test/Spring-test.js
nkbt: Nah... just nit.  No extra `{}` and `return`

test/Spring-test.js
bsansouci: Ok I'll bring them down to 5 (just to have 4 after the initial 0).

I am concerned by precision too. _Can_ it really vary though? Aren't those things deterministically wrong? I'd say truncating if anything.

test/mockRaf.js
bsansouci: Ha! turns out we don't need to set the value to something different than 0, we're only using it for delta time right now :D

test/Spring-test.js
bsansouci: Doesn't `push` return the length of the new array? Not sure how react will do with that.

test/Spring-test.js
nkbt: You can use larger numbers (10000) and math.round.

test/Spring-test.js
chenglou: can't return numbers from render.

test/Spring-test.js
chenglou: nit: `round` would actually make regression pass if previously it was 0.7 and now it's 0.8. Never gonna happen but still, no need to round I think. Plus, yes, it's deterministically wrong if ever, so we do want that.

nkbt: This is why I was suggesting to use large nums. Or you can use round with
precision alternatively.

chenglou: Oh the "and" confused me =). use large nums then

test/Spring-test.js
nkbt: `push(val) && null`? Nevermind, that looks worse then `return` :)

test/Spring-test.js
bsansouci: Ok so just interpolate between 0 and say 10000, and then round?

bsansouci: I'll let @chenglou change it later

chenglou:
This is great

PR #116, opened 2015-08-03
Fix React 0.14 compatibility for all beta

ludovicofischer:
See http://semver.npmjs.com/

I think it might also work for rc, but I am not sure.


threepointone:
Does this work? I believe you have to explicitly mark out each beta/rc version; not sure ^ will work. Happy if I'm wrong :)

threepointone:
ok, semver.npmjs seems to claim it will. looks good!

chenglou:
the more you know...

Issue #77, opened 2015-07-18
TransitionSpring bugs on willEnter


chenglou:
Repro: demo3, type `p`. Remove `p`. Original list doesn't come back because [this](https://github.com/chenglou/react-motion/blob/1ec99fef09a9510a643357f7e2b9c8b7f02e8e68/src/Spring.js#L361) is met.

chenglou:
@iclanzan

iclanzan:
Interesting, I never waited log enough for the animations to finish in my tests so I never ran into this. :)

How is it that it was working before though?

chenglou:
Probably through some funny combination of booleans?

iclanzan:
No, but I mean you were still checking the velocity to decide when to stop and yet it was working.

chenglou:
Aren't we talking about how to restart here?

iclanzan:
I guess that’s the real issue, yes.

iclanzan:
What do you think of changing [this line](https://github.com/chenglou/react-motion/blob/master/src/Spring.js#L353) to: `mergedValue[key] = endValue[key];`?

That seems to solve it.

chenglou:
I'll have to think about the implication of this difference for TransitionSpring. I _think_ this is fine. @bsansouci

bsansouci:
This will make the raf stop after the first render, which was handled by [this](https://github.com/chenglou/react-motion/blob/5541cca8d569661ec7e183414ca95118faaab875/src/Spring.js#L152) previously. Preventing certain CSS transitions to work. 
Example [here](https://gist.github.com/bsansouci/c8c760339f5134c50ac5).

A possible way to fix this is to change [this line](https://github.com/chenglou/react-motion/blob/master/src/Spring.js#L361) to:
```js
if (noVelocity(currVelocity) && noVelocity(newCurrVelocity)) {
  if (this.oneMoreTime) {
    this.unsubscribeAnimation();
    this.unsubscribeAnimation = undefined;
    this.oneMoreTime = false;
  } else {
    this.oneMoreTime = true;
  }
} else {
  this.oneMoreTime = false;
}
```

and initialize `this.onMoreTime` [here](https://github.com/chenglou/react-motion/blob/master/src/Spring.js#L281).

iclanzan:
[...]

PR #121, opened 2015-08-04
Add more tests.

bsansouci:
Here's a list of the added tests:
- parent kill child Spring
- TS unmount/mount correctly (and https://github.com/chenglou/react-motion/issues/77)
- init value + final value
- TS init value + final value

### DIFF
--- a/test/Spring-test.js
+++ b/test/Spring-test.js
@@ -78,7 +78,7 @@ describe('Spring', () => {
     });
     TestUtils.renderIntoDocument(<App />);
 
-    // Initial render
+    // Checking initial render
     expect(count).toEqual([0]);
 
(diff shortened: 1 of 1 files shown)


test/Spring-test.js
```diff
@@ -391,4 +391,239 @@ describe('Spring', () => {
...
+        }
+        return this.state.isAlive && <App kill={() => this.setState({isAlive: false})}/>;
+      },
+    });
+
+    TestUtils.renderIntoDocument(<Parent />);
+
+    expect(count).toEqual([0]);
+    mockRaf.manySteps(4);
+    shouldKill = false;
```
chenglou: ?

bsansouci: This is just testing for that one edge case of rendering something that has been unmounted.

PR #122, opened 2015-08-04
fix babel warning for demos/docs build

threepointone:
via the directions on https://babeljs.io/docs/usage/browser/, copied `browser.min.js` from `babel-core` into `demos/demo6` and included it directly into the page (alternately, it could be concatenated with `all.js`, but this is simpler.)


chenglou:
¯\_(ツ)_/¯

PR #123, opened 2015-08-04
Fix a bug in animationLoop when having nested Springs.

bsansouci:
This one was hard to track down. Say you have two springs one inside
the other, there's the `owner` and the `child`. They both mount, and
therefore call `startAnimation` which will register a `raf`.

The `child` steps first because it registered the `raf` callback first.
In the step it checks `noVelocity` and say in our case it's true, it'll
calls `stopAnimation`. `stopAnimation` will set `val.active = false`.
Then the `owner` steps.

We then loop through the `Spring`s, starting with the `child` and
render them. We render the `child` and check if `child.active` is true.
In our case it's not, so we don't push it in the array of future `raf`s.

We then render the `owner` which makes the `child` render and call
`startAnimation`. The call to `startAnimation` will set
`val.active = true`, but it's too late, we decided not to push that
Spring in the array of future `raf`s at the last tick of the loop.
That render is done we check the length of that array and see that it's
empty, so we stop.

The fix is to call all of the renders first, and then loop through
the `Spring`s and push in the future `raf`s array what needs to step.

This includes a test.

### DIFF
--- a/src/animationLoop.js
+++ b/src/animationLoop.js
@@ -58,14 +58,17 @@ export default function configAnimation(config = {}) {
     accumulatedTime = accumulatedTime - frameNumber * timeStep;
 
     // Render and filter in one iteration.
-    const newAnimRunning = [];
     const alpha = 1 + accumulatedTime / timeStep;
     for (let i = 0; i < animRunning.length; i++) {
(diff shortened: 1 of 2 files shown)


chenglou:
This is getting really heavy. I'll think about what to do with this.

PR #126, opened 2015-08-06
Update inject-loader to use 2.0.1 now it’s published

plasticine:
Hey there :)

I’m the author of inject-loader. This PR gets react-motion off the master branch of inject-loader and onto a published version (2.0.1 to be precise), which now includes the code that necessitated using master in the past


chenglou:
Woohoo! Thanks =)

PR #127, opened 2015-08-07
Fix unmounted state mutation

daniel-gustafsson:
TransitionSpring should likely have the same fix for unmounting that Spring element has regarding state mutation.

### DIFF
--- a/src/components.js
+++ b/src/components.js
@@ -231,6 +231,7 @@ export default function components(React) {
 
     componentWillUnmount() {
       this.stopAnimation();
+      this.hasUnmounted = true;
     },
 
     startAnimating() {


chenglou:
@bsansouci

bsansouci:
Good call!

bsansouci:
Everything works! Thanks a lot.

daniel-gustafsson:
Thanks for accepting it!

I wonder though if I made a mistake with the pull request, because I can't seem to find the actual commit in this repository anywhere. It was my very first pull request so apologies if I got it wrong.

chenglou:
They're here alright: https://github.com/chenglou/react-motion/commit/30a780a2a4e5bb739c7c24191b268f675898c436 and https://github.com/chenglou/react-motion/commit/589ae9b3851442a165deb321a6039665a8e1672f =)

PR #134, opened 2015-08-20
improve the doc to indicate how to get Spring

gre:
Hi,

First of all thanks for this great library.

as the first time using this library, the README did not help me to know how to get the Spring out of the library, it is not mentioned anywhere in the README and I think it is important as a "Getting Started".

The examples didn't help neither because they directly requires the Spring file (I think it should be better to extract the things from the library root).

So for now, maybe I'm wrong, but the only way to figure this out is to look at the library code.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![react-motion channel on slack](https://img.shields.io/badge/slack-react--motion%40reactiflux-61DAAA.svg?style=flat)](https://reactiflux.slack.com/messages/react-motion/)
 
 ```js
-const { Spring } = require("react-motion");
+const {Spring} = require("react-motion");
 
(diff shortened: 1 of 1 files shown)


chenglou:
It says before [this line](https://github.com/chenglou/react-motion/blob/56df77d44e490ed5d5fda678e445d6ee9f2cc3da/README.md#spring-) that the library exports `Spring`, among others.

That being said, I guess not everyone sees that. Can you remove the space between `{}`? I'll merge this after.

gre:
done :)
I've used your neat library for a cool example I'm going to release soon, I'll show you ;)

chenglou:
Thanks!

Issue #137, opened 2015-08-29
Is it possible to animate the mounting of a component?


wilfreddenton:
For example, in this [demo](https://cdn.rawgit.com/chenglou/react-motion/e8f42dcd9678a8cea8648a3cf4f994583a99e7f7/demos/demo3/index.html), is it possible to have the listview animate when it's mounted in the same way that searching "asdf" and then "" animates the view? This would be useful for when a listview is toggleable.

threepointone:
Total guess, but should be possible by starting with 0 items, and then adding them in after a setTimeout(fn, 0)?

souporserious:
Yes, just provide a defaultValue and then whatever you want your endValue to be. You can see a working example here: http://codepen.io/souporserious/pen/dde7064750f523bfa934da2d6e201af4?editors=001

chenglou:
^ Yep!

PR #138, opened 2015-08-30
updated todos to show mounting animation

souporserious:
In response to this issue https://github.com/chenglou/react-motion/issues/137

### DIFF
--- a/demos/demo3/Demo.jsx
+++ b/demos/demo3/Demo.jsx
@@ -76,7 +76,7 @@ const Demo = React.createClass({
   },
 
   // actual animation-related logic
-  getEndValue() {
+  getEndValue(prevValues) {
     const {todos, value, selected} = this.state;
     return Object.keys(todos)
       .filter(date => {
@@ -88,7 +88,7 @@ const Demo = React.createClass({
       })
       .reduce((configs, date) => {
         configs[date] = {
-          height: {val: 60, config: presets.gentle},
+          height: {val: prevValues ? 60 : 0, config: presets.gentle},
           opacity: {val: 1, config: presets.gentle},
           data: todos[date],
         };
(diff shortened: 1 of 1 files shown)


chenglou:
You could use `defaultValue`

souporserious:
Thought it would be easier like that. Want me to update it using `defaultValue`?

chenglou:
Yeah please do. using `prevValue` with a null check isn't intuitive. Otherwise I wouldn't have exposed `defaultValue`

chenglou:
:+1:  thanks!

wilfreddenton:
@souporserious with this commit should demo3 now animate in the todo items when the page loads/component mounts? I cloned the repo and ran it locally to see how to get the functionality in my project, but it's not animating for me.

chenglou:
That's weird, it animates for me. Does your console say anything?

souporserious:
Yeah works fine for me as well. You did `npm start` right?

wilfreddenton:
Doh! Idk why I thought the demos would come transpiled. It's working for me. Thanks. I'm curious though, is it possible to also have unmounting animations? Like if there was some boolean that mounted and unmounted the list view.

chenglou:
The demo actually does unmounting animation, when you filter through it or remove one

wilfreddenton:
Right, it handles the unmounting of the individual list items. I was wondering if it could handle the unmounting of the containing unordered list but I see now that it's a better idea to just remove the items instead of the whole view when I want the list to disappear. This way the mounting/unmounting is all handled by react-motion. Thanks @chenglou and @souporserious so much for taking the time to help me!

chenglou:
Right, there's no magic render-that-keeps-child. I've put this as a FAQ in the readme since it's not the first time I've encountered this question. Thanks!

PR #145, opened 2015-09-02
eslint fixes

threepointone:
fixes these lint errors, fixes #144 
```
demos/demo6/babel.js
  0:0  error  Maximum call stack size exceeded

native/index.js
   1:0   error  Strict mode is not permitted                                     strict
   7:0   error  Unexpected var, use let or const instead                         no-var
   7:0   error  All "var" declarations must be at the top of the function scope  vars-on-top
   9:0   error  All "var" declarations must be at the top of the function scope  vars-on-top
   9:0   error  Unexpected var, use let or const instead                         no-var
  11:0   error  All "var" declarations must be at the top of the function scope  vars-on-top
  11:0   error  Unexpected var, use let or const instead                         no-var
  13:0   error  All "var" declarations must be at the top of the function scope  vars-on-top
  13:0   error  Unexpected var, use let or const instead                         no-var
  15:0   error  Unexpected var, use let or const instead                         no-var
  15:0   error  All "var" declarations must be at the top of the function scope  vars-on-top
  17:0   error  Unexpected var, use let or const instead                         no-var
  17:0   error  All "var" declarations must be at the top of the function scope  vars-on-top
  19:0   error  Unexpected var, use let or const instead                         no-var
  19:0   error  All "var" declarations must be at the top of the function scope  vars-on-top
  19:18  error  ["default"] is better written in dot notation                    dot-notation
  19:45  error  ["default"] is better written in dot notation                    dot-notation
  21:0   error  All "var" declarations must be at the top of the function scope  vars-on-top
  21:0   error  Unexpected var, use let or const instead                         no-var
  22:0   error  Unexpected var, use let or const instead                         no-var
  22:0   error  All "var" declarations must be at the top of the function scope  vars-on-top
  26:0   error  Unexpected var, use let or const instead                         no-var
  26:0   error  All "var" declarations must be at the top of the function scope  vars-on-top
  28:0   error  Unexpected var, use let or const instead                         no-var
  28:0   error  All "var" declarations must be at the top of the function scope  vars-on-top
  30:18  error  ["default"] is better written in dot notation                    dot-notation
  31:0   error  Unexpected var, use let or const instead                         no-var
  31:0   error  All "var" declarations must be at the top of the function scope  vars-on-top
  32:15  error  ["default"] is better written in dot notation                    dot-notation
  32:42  error  Missing trailing comma                                           comma-dangle

test/Spring-test.js
  354:8  error  Do not use setState in componentDidMount  react/no-did-mount-set-state

[...]

PR #146, opened 2015-09-02
lint fix : demo - demos

threepointone:
missing from #145

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -8,4 +8,4 @@ server.js
 karma.*.js
 test/integration
 native/index.js
-demo/demo6/babel.js
+demos/demo6/babel.js


chenglou:
Does this work? Just tested locally and it still doesn't work.

threepointone:
Works for me?

threepointone:
What error are you still getting?

chenglou:
Did you try a fresh npm install? I think it's eslint again
```
.../react-motion/node_modules/eslint-plugin-react/lib/rules/prop-types.js:450
          propTypes.parent.type !== 'AssignmentExpression' &&
                          ^
TypeError: Cannot read property 'type' of undefined
```

threepointone:
working just fine with a fresh npm install

chenglou:
Humm that's weird. Switched to another laptop and now it works. Probably because of the artifacts of the `integration/` stuff which isn't tracked by this repo.

PR #148, opened 2015-09-03
set this.hasUnmounted to false in componentWillUnmount method of the TransitionSpring component

wilfreddenton:
I was working on a component that uses react-motion and noticed that it caused react to throw the setState warning when I went to a new route. I located the source of this issue in `src/components.js` in TransitionSpring's `animationRender` method. The Spring component sets `this.hasUnmounted` to `false` in `componentWillUnmount` but TransitionSpring does not. When `animationRender` checks `!this.hasUnmounted` it evaluates to true and the setState code is run causing react to throw the warning. So I added `this.hasUnmounted = true;` to `componentWillUnmount` in TransitionSpring. The warning is no longer thrown.

### DIFF
--- a/src/components.js
+++ b/src/components.js
@@ -352,6 +352,7 @@ for the upgrade path. Thank you!`
 
     componentWillUnmount() {
       this.stopAnimation();
+      this.hasUnmounted = true;
     },
 
     startAnimating() {


chenglou:
@bsansouci

bsansouci:
Thanks! We had already fixed this but we might've overwritten that change. I think we can merge @chenglou.

chenglou:
Did we ever add it for TS? A quick search though history says no. Anyway, @bsansouci can you add a test against this? I'll let you merge this before/after.

chenglou:
Alright, merging this. @bsansouci I'll try to fix all the tests today so that you can test again this. Thanks @wilfreddenton!

wilfreddenton:
@chenglou no problem! Thanks for all the work you've put in on this project. It's really awesome.

PR #152, opened 2015-09-12
Update React peerDependency to rc1

danieldunderfelt:
To enable smooth usage of React 0.14 RC1 without NPM errors on install.


chenglou:
Thanks!

PR #173, opened 2015-09-29
Add missing jasmine dependency.

ludovicofischer:
Hi, when I tried to run `npm test`, it failed at first because the `jasmine-core` dependency was missing. Per karma-jasmine documentation (https://github.com/karma-runner/karma-jasmine) since 0.3.0 you need to add an explicit `jasmine-core` dependency since karma-jasmine does not pull it any more.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "eslint-plugin-react": "^2.7.0",
     "inject-loader": "^2.0.1",
     "isparta-loader": "^0.2.0",
+    "jasmine-core": "^2.3.4",
     "karma": "^0.12.37",
     "karma-coverage": "^0.4.2",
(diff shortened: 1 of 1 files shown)


chenglou:
That's weird. Are you sure? I've removed node_modules and reinstalled and everything worked fine still.

ludovicofischer:
Hi, are you sure you don’t have a global jasmine install (maybe from the jasmine command line tool). I also started from empty `node_modules` (I had just cloned the repository) and the tests would not run because of the missing dependency.

nkbt:
jasmine core is needed, it was probably installed as peer-dep and now with npm@3 it is not.

chenglou:
Ah. I'm still on npm 2. Thanks!

PR #174, opened 2015-09-29
Fix the build it yourself one-liner

wyze:
It was missing the change directory step.
It was trying to run `npm i` && `npm run dev` on parent directory.


chenglou:
Oops, thanks!

PR #176, opened 2015-09-30
Unminified examples, fixes #170

### LINKED ISSUE #170: Unminified example code
It'd be awesome if the examples linked to from the `readme.md` were unminified so one could easily see how such examples were accomplished.

nkbt:
I extracted React from demos, so files are much much lighter and easier to debug

Minified react is copied before build/start from node_modules. So it is always up-to-date.


nkbt:
The only large demo is 6, since there are more large vendor modules used, I decided not to mess with them for now.

chenglou:
demo 6 is useless right now, it's fine

chenglou:
React is minified but https://github.com/nkbt/react-animation/blob/unminified-examples/demos/demo8/all.js#L756 isn't (I use the trick inside RM too)
This isn't a problem because I don't use it elsewhere, for now.

PR #181, opened 2015-10-02
Handle `undefined` and `null` in style objects

### LINKED ISSUE #180: Uncaught TypeError: Cannot read property 'config' of undefined
I am upgrading from 0.2.7 to 0.3.0 and I ran into an error:

> Uncaught TypeError: Cannot read property 'config' of undefined

This happens when I am using a `TransitionMotion` with a style object that has a key who's value is `undefined`, e.g.:

```js
{
  opacity: spring(0),
  extraText: undefined,
}
```

This is happening on https://github.com/chenglou/react-motion/blob/d478fe6ba1a5ba74ab364747f063e16fa8c3185f/src/updateTree.js#L38

I can side-step the issue by setting that value to something other than `undefined` or `null`, but it would be nice if I didn't have to. I believe this could be avoided by adding a couple of checks here and there, for example:

```js
if (!style[key].config) {
```

could be changed to

```js
if (!style[key] || !style[key].config) {
```

lencioni:
I am upgrading from 0.2.7 to 0.3.0 and I ran into an error:

> Uncaught TypeError: Cannot read property 'config' of undefined

This happens when I am using a `TransitionMotion` with a style object
that has a key who's value is `undefined`, e.g.:

  {
    opacity: spring(0),
    extraText: undefined,
  }

To make react-motion easier to work with, I have added guards in a
number of places.

Fixes #180.

### DIFF
--- a/test/stripStyle-test.js
+++ b/test/stripStyle-test.js
@@ -1,0 +1,7 @@
+import stripStyle from '../src/stripStyle';
+
+describe('stripStyle', () => {
+  it('should ignore `undefined`', () => {
+    expect(stripStyle({a: undefined})).toEqual({a: undefined});
+  });
+});
--- a/test/updateTree-test.js
+++ b/test/updateTree-test.js
@@ -1,7 +1,15 @@
-import {updateCurrentStyle, updateCurrentVelocity} from '../src/updateTree';
+import {interpolateValue, updateCurrentStyle, updateCurrentVelocity} from '../src/updateTree';
 import {spring} from '../src/react-motion';
 
 const FRAME_RATE = 1 / 60;
+
+describe('interpolateValue', () => {
+  it('should handle `undefined`', () => {
+    const nextStyle = {a: undefined};
+    const prevStyle = {a: undefined};
+    expect(interpolateValue(1, nextStyle, prevStyle)).toEqual({a: undefined});
+  });
+});
 
 describe('updateCurrentStyle', () => {
   it('should jump to style when there is no config', () => {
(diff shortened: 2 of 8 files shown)


chenglou:
These checks should be `x == null` to avoid catching 0

lencioni:
That's a fair point. I'll update my PR shortly.

test/stripStyle-test.js
```diff
@@ -0,0 +1,15 @@
+import stripStyle from '../src/stripStyle';
```
chenglou: Neat, though can you pull this out into another PR once I merge this one?

lencioni: Should I just include the test for ignoring `undefined` in this PR and put the other 2 tests in a new PR?

chenglou: That works too, yeah.

lencioni:
New version is ready to go. After you merge I'll submit a new PR for the additional tests.

test/updateTree-test.js
```diff
@@ -65,6 +81,14 @@ describe('updateCurrentVelocity', () => {
       .toEqual({a: 0, b: 20.4});
   });
 
+  it('should handle `undefined`', () => {
+    const currentStyle = {a: undefined};
+    const currentVelocity = {a: undefined};
+    const style = {a: undefined};
+    expect(updateCurrentVelocity(FRAME_RATE, currentStyle, currentVelocity, style))
+      .toEqual({a: 0});
```
chenglou: Humm, on second thought this should probably be `undefined`. My fault. Will reconsider this.

chenglou:
Cool, done.

test/updateTree-test.js
lencioni: Yeah, I found this to be a little unexpected. :)

lencioni:
My pleasure.

PR #183, opened 2015-10-02
Add some tests for stripStyle()

lencioni:
While fixing #180, I noticed that stripStyle() was untested, so I
decided to add a couple of basic tests here.

### DIFF
--- a/test/stripStyle-test.js
+++ b/test/stripStyle-test.js
@@ -1,6 +1,15 @@
 import stripStyle from '../src/stripStyle';
+import spring from '../src/spring';
 
 describe('stripStyle', () => {
+  it('should return spring object into value', () => {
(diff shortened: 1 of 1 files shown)


test/stripStyle-test.js
```diff
@@ -1,6 +1,14 @@
 import stripStyle from '../src/stripStyle';
 
 describe('stripStyle', () => {
+  it('should return spring object into value', () => {
+    expect(stripStyle({a: {val: 1, config: [1, 2]}})).toEqual({a: 1});
```
chenglou: can you change these to the opaque `spring(1, [1, 2])`? I'll update the rest of the codebase to follow this.

lencioni: Done!

chenglou:
Thanks again.

PR #188, opened 2015-10-04
Update karma and related libraries to precent warning on Node >= 4.0

ludovicofischer:
I read the changelog and it looks like there are no changes impacting the end user.


chenglou:
Thanks!

PR #193, opened 2015-10-06
Windows-compatible npm-scripts, fixes #191

nkbt:
Some minor changes to npm scripts and configs so they all fully run under Windows.

PR #194, opened 2015-10-06
Add pretty diff jasmine matcher

nkbt:
We can adjust the way it looks in `prettyMatcher.js`. Now this matcher overrides `isEqual`, but it is trivial to override others if necessary

I had to remove `nyan` reporter, since it messes up output, I tried different settings, didn't help much.

I also tried other pretty-printing solutions:
- https://www.npmjs.com/package/karma-jasmine-diff-reporter
- https://www.npmjs.com/package/jasmine-json
- and some other random suggestions from SO, aso

Either way they were all broken or not good enough, so, you know, wrote my own, `¯\_(ツ)_/¯`

It supports circular objects too `\m/` because one never knows...


nkbt:
There could be some conflicts with #193

chenglou:
yeah can you rebase please?

nkbt:
Sure. I may later update jasmine-diff-reporter so when it is released, most of this code might be removed.

nkbt:
Rebased, fixed conflicts

chenglou:
So what's the story with jasmine-diff-reporter? Will it support this?

nkbt:
@chenglou there are some conceptual issues implementing it on reporter level. Author is working on that and I will help him as much I can. So potentially if it is working well, I'll file another PR, but meanwhile current solution works pretty well.

Issue #190, opened 2015-10-05
Rename Demo Folders to What They Do


jclyons:
Can you rename the demo directories to what each one is? It'd be great if you could label them by what they do so it's easier to tell the difference. I reference them a lot when I'm working and I usually have to cross-reference with the README.md. Small nitpick - thanks for the great library.

chenglou:
Been on my mind for a while too. Done!

PR #195, opened 2015-10-06
Rename demo folders

stevoland:
To include description as per: https://github.com/chenglou/react-motion/issues/190


chenglou:
Thanks!

nkbt:
There is a problem with this renaming  #197

PR #199, opened 2015-10-07
Better .npmignore

nkbt:
A little cleanup, updated with some new files

NOTE: I allowed sources to be published on NPM. So it is possible to have react-motion as part of end-project build. With all sourcemaps and so on (simply by doing `import {spring} from 'react-motion/src/react-motion'` or better using webpack module alias instead)


nkbt:
I vote for that day to be today ;)

nkbt:
And yes, flow is a perfect pro-argument here too.

PR #201, opened 2015-10-07
Publish to GitHub Pages, fixes #24

### LINKED ISSUE #24: GitHub Pages for demos
It'd be slick if the demos were available over GH Pages. Then people wouldn't need to clone the repo just to see the demos. =)

Should be as easy as pushing to a `gh-pages` branch, but I might be overlooking something.

Nice work!

nkbt:
No more pre-built demos! Development works as usual, hot reloading and things...

Bonus: fixed up demo6 so it does not have any extra assets that we can pull from node_modules. Similar thing with css should be done to other demos. And when all css is abstracted we can remove `index.html` too and generate it on-the-fly with `html-webpack-plugin` instead.


nkbt:
Actually, http://chenglou.me/react-motion/demos/demo0-simple-transition is already available since I pushed `gh-pages` from current master. But it needs some changes there after this PR is merged in.

PR #209, opened 2015-10-09
Restructure API docs to use better type signatures

bjz:
I was having a great deal of trouble understanding the API, so I've attempted to make the type signatures much clearer in the docs. I may have got some things wrong though. I originally stuck to the Haskell-esque syntax for the type signatures, but opted to switch to TypeScript syntax because it allows you to annotate the function parameters. I hope this helps!

[Rendered](https://github.com/bjz/react-motion/blob/d072fa928baf5a5955b9ae1bc182e111377ca791/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ let config = spring(10, [120, 17]);
 
 This is not the `SpringConfig` component in version <0.3.0 - this has been renamed to `Motion` (see below).
 
-### `class Motion extends React.Component`
+### `<Motion />`
 
 #### `props`
 
@@ -120,7 +120,7 @@ Required, which is passed an interpolated style object, e.g. `{x: 5.2, y: 12.1}`
 </Motion>
 ```
 
-### `class StaggeredMotion extends React.Component`
+### `<StaggeredMotion />`
 When you want to animate a list of items, you can certainly create an array of `Motion`s and animate each. However, you often want to "stagger" them, i.e. animate items in one after another with a delay. Hard-coding this duration goes against the very purpose of spring physics. Instead, here's a natural, physics-based alternative, where "the destination position of an item depends on the current position of another".
 
 #### `props`
@@ -154,7 +154,7 @@ Required. Similar to `Motion`'s `children`, but accepts the array of interpolate
 </StaggeredMotion>
 ```
 
-### `class TransitionMotion extends React.Component`
+### `<TransitionMotion />`
 **The magic component that helps you to do mounting and unmounting animation.** Unlike React's `TransitionGroup`, instead of retaining a few items in a list when they disappear, `TransitionMotion` diffs on the shape of its `styles` object prop.
 
 **The general idea**
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ Optional. The value when the component first renders (ignored in subsequent rend
 ##### `style: Style`
 Required. Must have the same keys throughout component's existence. Must have the same keys as `defaultStyle` (if provided). Similar to `defaultStyle`, but asks for a `spring` configuration as the destination value: `{x: spring(10), y: spring(20, [120, 17])}`.
 
-##### `children: (interpolated: Style) => ?React.Component`
+##### `children: (Style) => ?React.Component`
 Required, which is passed an interpolated style object, e.g. `{x: 5.2, y: 12.1}`. Must returns a React element to render.
 
 ```jsx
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,25 +52,67 @@ For 95% of use-cases of animating components, we don't have to resort to using h
...
+
+### `class Motion extends React.Component`
+
+#### `props`
+
+##### `defaultStyle?: Style`
 Optional. The value when the component first renders (ignored in subsequent renders). Accepts an object of arbitrary keys, mapping to initial values you want to animate, e.g. `{x: 0, y: 10}`.
 
-#### `style: Object`
+##### `style: Style`
```
nkbt: :+1: nice

nkbt:
We are now moving to Flow code annotations, so it would make much more sense to use those for docs too instead of TypeScript

See #187

nkbt:
In general LGTM, TS-->Flow annotations should be a quick fix

bjz:
Oh cool, that is most excellent. Can you gnerate docs from Flow, like with typescript?

bjz:
I'll amend with the Flow syntax, but it would be even better if they were auto-generated.

nkbt:
I don't think there was any intention to auto-generate docs from Flow. It does not really have any context similar to what JSDoc does (via comments).

bjz:
Ah, that's a shame. You can get very nice dos with little effort using generators like [TypeDoc](http://typedoc.io/), automatically generating hotlinked type signatures.

bjz:
Wasn't much of a change to use flow-style annotations. Just meant I could use `?React.Component` as opposed ti `(React.Component | void)`

nkbt:
:+1: LGTM, @chenglou there are some noticeable text changes, do they capture everything right?

bjz:
Any updates on this?

chenglou:
(Sorry catching up to like, everything... Will check more closely tonight)

README.md
```diff
@@ -52,25 +52,67 @@ For 95% of use-cases of animating components, we don't have to resort to using h
 This library also provides an alternative, more powerful API for React's `TransitionGroup`.
 
 ## API
-The library exports `Motion`, `StaggeredMotion`, `TransitionMotion`, `presets`, `spring` and `utils`.
 
-#### `spring: number -> ?[stiffness, damping] -> ConfigurationObject`
-(**Note**: not the `Spring` component in version <0.3.0.)
-The pervasive helper used to specify the how to animate to the destination value, e.g. `spring(10, [120, 17])` returns an opaque configuration that describes "an animation to the value 10, with a physics spring's stiffness of 120 and damping of 17". `spring(10)` without the spring configuration array defaults to `[170, 26]`. See below for more usage and see [here](#presets) for a list of convenient configurations the library exports.
+### `type PhysicsConfig = [number, number]`
```
chenglou: Really not sure about this. If you check the final effect: https://github.com/bjz/react-motion/blob/restructure-readme/README.md#api it's a lot of ceremony before some actual docs

README.md
```diff
@@ -52,25 +52,67 @@ For 95% of use-cases of animating components, we don't have to resort to using h
...
+
+```jsx
+let config = spring(10, [120, 17]);
+```
+
+#### Note
+
+This is not the `SpringConfig` component in version <0.3.0 - this has been renamed to `Motion` (see below).
+
+### `class Motion extends React.Component`
```
chenglou: `<Motion />` was much clearer

README.md
```diff
@@ -213,10 +262,22 @@ const Demo = React.createClass({
 });
 ```
 
-### `presets`
+### Modules
+
+#### `presets`
```
chenglou: Not sure about this. Seems like the link's helpful enough.

chenglou:
We're using Flow's type syntax. Also this is getting hard to read: 
screenshot 2015-10-14 01 27 49

The vague type annotations, where it's the case, are mostly intentional. I've debated long for/against this but ultimately went with a general type + a concise description. Formalizing things like this makes everything much more rigorous but actually decreases readability imo...

chenglou:
Hey @bjz, this is really late, but hope you like the [new docs](https://github.com/chenglou/react-motion/blob/6a7a0f3f72610c6a7fdad2920b44e728ca6e172d/README.md#api)! =) 

Proper Flow type signature, with an exposed public types file.

Sorry that this took so long. To be fair, Flow wasn't (and is still a bit buggy sometimes) ready back then either. This PR has diverged too much so I'll close it. But thanks a lot for the initiative!

bjz:
No worries - it looks great now!

PR #217, opened 2015-10-13
Ignore sublime configuration files

operatino:
Noticed that latest NPM package release (0.3.0) contains Sublime configuration files, which seems like got there accidentally. To prevent it further, it's better to ignore those files from adding to NPM package.


chenglou:
Damn it! I thought I made sure this didn't happen again...
Thanks!

PR #238, opened 2015-11-10
Switch to compare x with rest on errorMargin check

istarkov:
Looks like it's better to check x with rest value, not previous.
For example if errorMargin becomes a user parameter, there will situations with small velocity and small dx values but x is far from the rest.

### DIFF
--- a/src/stepper.js
+++ b/src/stepper.js
@@ -25,7 +25,7 @@ export default function stepper(
   const newV = v + a * frameRate;
   const newX = x + newV * frameRate;
 
-  if (Math.abs(newV - v) < errorMargin && Math.abs(newX - x) < errorMargin) {
+  if (Math.abs(newV - v) < errorMargin && Math.abs(newX - destX) < errorMargin) {
     return [destX, 0];
   }
 
--- a/src/stepper.js
+++ b/src/stepper.js
@@ -1,5 +1,5 @@
 /* @flow */
-const errorMargin = 0.0001;
+const errorMargin = 0.001;
 
 export default function stepper(
   frameRate: number,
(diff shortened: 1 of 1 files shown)


istarkov:
Also looks like it's better  for spring animation (with check added above) to check velocity
`Math.abs(newV) < errorMargin` but not an acceleration `Math.abs(newV - v) < errorMargin`, and it allows to increase errorMargin Look at second commit

istarkov:
PS: Without ability to write custom stepper or set errorMargin as spring parameter there are a lot of small visually bad animation movements.

chenglou:
The logic here is tricky. Does this glitch with animations whose destination is 0? What about the edge-cases of TransitionMotion where you're adding such an item, etc. (this is super vague but can you see what I mean?)
Also, how relevant is this if we implement custom error margin?

istarkov:
1) Destination value can be zero or any other value, this part `Math.abs(newX - destX)` contains delta not destination itself so it can be any numeric value.

2) It's more easy to user think in terms like - `if velocity is small enough and distance to destination is small we can stop animation` instead of current `if acceleration is small and velocity is small we can stop animation`

3) For a big errorMargin master solution could stop animation at first step even if destination value is far far away from value because of acceleration and velocity is less then errorMargin.
Solution proposed above will not stop.

4) I've just made a search over github and fast looking into the code gives
Looks like [facebook rebound use the same approach](https://github.com/facebook/rebound-js/blob/master/rebound.js#L754) as above
Looks like [JNWSpringAnimation same approach](https://github.com/jwilling/JNWSpringAnimation/blob/master/JNWSpringAnimation.m#L263) use the same approach (be sure that proposedValue - previousValue is Velocity) or similar

5) I'm not an expert in `edge-cases of TransitionMotion` so can't say nothing about, need to read more.

chenglou:
You know what, this is probably right. If TransitionMotion screws up after this then we should fix that instead.

Sorry for the delay! Thanks =).

istarkov:
Thank you for this library!

PR #242, opened 2015-11-17
Use npmcdn for Bower and CDN

nkbt:
Since it is not yet published, I tested on react itself and it works well.

```sh
[nkbt@nkbt] tmp $ bower install https://npmcdn.com/react@0.14.2/dist/react.min.js -S
bower react.min#*               cached https://npmcdn.com/react@0.14.2/dist/react.min.js
bower react.min#*             validate https://npmcdn.com/react@0.14.2/dist/react.min.js#*
bower react.min#*              install react.min

react.min bower_components/react.min

[nkbt@nkbt] tmp $ cat bower_components/react.min/.bower.json
{
  "name": "react.min",
  "_cacheHeaders": {
    "Content-Length": "135543",
    "Content-Type": "application/javascript; charset=utf-8"
  },
  "main": "index.js",
  "_source": "https://npmcdn.com/react@0.14.2/dist/react.min.js",
  "_target": "*",
  "_originalSource": "https://npmcdn.com/react@0.14.2/dist/react.min.js"
}

[nkbt@nkbt] tmp $ cat bower.json
{
  "name": "tmp",
  "version": "0.0.0",
  "authors": [
    "Nik Butenko <nik@butenko.me>"
  ],
  "moduleType": [
    "node"
  ],
  "license": "MIT",
  "ignore": [
    "**/.*",
    "node_modules",
    "bower_components",
    "test",
    "tests"
  ],
  "dependencies": {
    "react.min": "https://npmcdn.com/react@0.14.2/dist/react.min.js"
  }
}
```

So we can remove build from git :birthday:


chenglou:
o tldr: no more need for bower builds checked in ever, anymore? Seems like I could just press the merge button but this is too good to be true

nkbt:
ahaha yeah, I think so.

The only downside for now - no sourcemap. Only one build file. Later I am going to publish `zip` file with all pre-compiled stuff (like, styles, multiple JS files, source-maps), bower supports zips (afaik). But the first step is this PR and it works well as for my tests.

You can try that by yourself with reactjs.

I am not sure how to test this stuff with rails-bower (or whatever those rails people use), since I do not do rails. If anyone can confirm it uses just a normal bower.json and supports plain URLs, then we are completely good ditching this awkward bower stuff.

chenglou:
I'd like to have a magical npmcdn that also bundles the npm module into one script consumable by bower.

Or not have bower.

nkbt:
That magic is too project-specific, I am pretty sure it is almost impossible to do properly. This is basically your webpack.prod config that does all the magic

chenglou:
(I was half joking btw =))

nkbt:
Some people may read the other half 

magic

PR #244, opened 2015-11-25
Add MotionBoundary / externalRender which lets you skip unnecessary renders

avi:
If the reason the subtree under the motion is rendered just because of the animation, stop rendering in the MotionBoundary, this significantly improves frame rate.
Added to Todo sample

### DIFF
--- a/demos/demo3-todomvc-list-transition/Demo.jsx
+++ b/demos/demo3-todomvc-list-transition/Demo.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {TransitionMotion, spring} from '../../src/react-motion';
+import {TransitionMotion, spring, MotionBoundary} from '../../src/react-motion';
 import presets from '../../src/presets';
 
 const Demo = React.createClass({
@@ -145,13 +145,14 @@ const Demo = React.createClass({
           <input className="toggle-all" type="checkbox" onChange={this.handleToggleAll} />
           <TransitionMotion defaultStyles={this.getDefaultValue()} styles={this.getEndValue()} willLeave={this.willLeave}
             willEnter={this.willEnter}>
-            {configs =>
+            {(configs, externalRender) =>
               <ul className="todo-list">
                 {Object.keys(configs).map(date => {
                   const config = configs[date];
                   const {data: {isDone, text}, ...style} = config;
                   return (
                     <li key={date} style={style} className={isDone ? 'completed' : ''}>
+                      <MotionBoundary externalRender={externalRender}>
                       <div className="view">
                         <input
                           className="toggle"
@@ -165,6 +166,7 @@ const Demo = React.createClass({
                           onClick={this.handleDestroy.bind(null, date)}
                         />
                       </div>
+                      </MotionBoundary>
                     </li>
                   );
                 })}
--- a/src/components.js
+++ b/src/components.js
@@ -68,11 +68,21 @@ export default function components(React) {
       };
     },
 
+    componentWillMount() {
+      this.externalRender = true;
+    },
+
+    componentDidUpdate() {
+      this.externalRender = false;
+    },
+
     componentDidMount() {
+      this.externalRender = false;
       this.startAnimating();
     },
 
     componentWillReceiveProps() {
+      this.externalRender = true;
       this.startAnimating();
     },
 
@@ -136,7 +146,7 @@ export default function components(React) {
 
     render() {
       const strippedStyle = stripStyle(this.state.currentStyle);
-      const renderedChildren = this.props.children(strippedStyle);
+      const renderedChildren = this.props.children(strippedStyle, this.externalRender);
       return renderedChildren && React.Children.only(renderedChildren);
     },
   });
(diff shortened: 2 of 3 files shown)


chenglou:
Were you looking for something like this? https://github.com/reactjs/react-static-container

avi:
Not at all what I wanted.
The idea isn't that the content is static, the idea is that the content
isn't aware of the animation and thus can only be rendered when motion
renders not because of it's internal setState but because it's parent
rendered.

The content inside the MotionBoundary is alive and will render every time
the parent of the Motion renders, this only allows you to save the cost of
rendering/reconciling the inner content on each render frame, it is
especially useful when the content is a huge react component for which we
can't add shouldComponentUpdate intelligently.

avi:
Just to clarify, the component itself does basically the same thing as the
MotionBoundary, the interesting part of this pull request is the addition
of the second parameter to the children render callback after the current
value which lets you provide the correct value for
shouldUpdate/externalRender.

bsansouci:
Hey @avi, this is cool because that should prevent nested Motions from rendering twice at every tick right? If I have a `Motion` inside another `Motion`, the inner one doesn't get rendered when the parent `Motion` gets rendered.
Would it make sense to have that on `Motion`/`TransitionMotion` directly?

avi:
Yes it would solve the double rendering and all we need to add to the components is the shouldComponentUpdate function...

appsforartists:
This sounds like a useful optimization.

Have you considered exposing it as a context variable and a mixin for the children?  Manually wrapping every Motion tree in a MotionBoundary and threading through the externalRender argument sounds like a lot of needless boilerplate for the primary use case.  Simply being able to mix in a MotionBoundary the same way you'd mix in PureRenderMixin would be a nicer API.

chenglou:
(For the record, I'm interested in this. Just haven't gotten around looking at this since I was doing the refactoring)

appsforartists:
On second thought, maybe a wrapping component is the way to go, so you can have children that don't know they are being animated.  Separation of concerns is usually a good thing.  :smiley:

appsforartists:
I just had an interesting conversation with @bsansouci:

We could create a `SimpleMotion` extension that handles the common case; something like this:

```javascript
var SimpleMotion = React.createClass(
  {
    shouldComponentUpdate: function () {
      return !this.context.inMotionRender;
    },

    getDefaultProps () {
      return {
        tagName: "div",
        translateUnits: "px",
        scaleUnits: "%",
        rotateUnits: "rad",
      };
    },

    render () {
      var {
        tagName,
        translateUnits,
        scaleUnits,
        rotateUnits,
        style: propsStyle,
        ...propsPassthrough
      } = this.props;

      return (
        <Motion
          { ...propsPassthrough }
        >
          {
            style => (
              React.createElement(
                tagName,
                {
                  style: {
                    willChange: 'transform',
                    transform: `translate3d(${ style.x }${ translateUnits }, ${ style.y }${ translateUnits }, ${ style.z }${ translateUnits }) 
                                scale(${ style.scale }${ scaleUnits }) 
                                rotate(${ style.rotate }${ rotateUnits })`,
                    ...propsStyle 
                  }
                },
                this.props.children
              )
            )
          }
        </Motion>
    }
  }
);
```

If all you are changing its x, y, rotation, and scale, it would give you a simple way to do that without having to think about child functions or external renders - it just looks like any other React component and optimizes `shouldComponentUpdate` to avoid thrashing `render`.  `Motion` et. al. would continue to exist for advanced cases.

It doesn't even have to live in react-motion; could be its own repo.

Thoughts?

bsansouci:
@appsforartists @avi Looking more at `react-static-container` it seems like it's very close to what you're saying. Specifically looking at slide number 16/49 [here](https://speakerdeck.com/vjeux/react-rally-animated-react-performance-toolbox), we can imagine the `Motion` passing down `externalRender` to the `shouldUpdate` prop instead of, in this example, `this.state.isAnimating`.

appsforartists:
@bsansouci Good find.  Looks like the same thing as `MotionBoundary` in the original PR, though you'd need to track `shouldUpdate` in userland without the other half of this PR that exposes it in `render`.  However, [now that we have `onRest`](https://github.com/chenglou/react-motion/pull/273), it is doable.

I'm still curious about making `isAnimating` a context variable so you can have something like `SimpleMotion` to do the fast thing by default.

avi:
I dislike limiting the power of motions to style/transform although that is indeed the most common use case, how about we add a singleLevel boolean prop which will inject the boundary into the result of the motion's children function response. This will let you use anything you want as the direct descendant of the motion without losing expressiveness and without adding cruft to the render function you supply.

appsforartists:
I don't think only those properties should be animatable, just that maybe there's an opportunity here to pave a cow path and make the thing most people want to do also the easiest thing to do.  Motion as it is should definitely still exist.  (Some of my favorite uses of Motion, for instance, animate a radius rather than a CSS property.)

I don't understand how `singleLevel` would work.  My understanding is that anything wrapped in a `StaticContainer` would reject updates and thus be frozen in its initial state.  Therefore, you'd have to look inside whatever `children` returns and inject a `StaticContainer` between the top-level element and its descendants.  That feels needlessly intrusive, and presumes that you are animating the wrapper component (which seems even more presumptive than saying you are animating GPU-acceleratable transform properties).

As skeptical as I am of argument sprawl, it's probably OK to add an `isAnimating` flag to the `children` signature.  However, I do wonder if there's also value in doing it in context and providing components like `MotionBoundary` or `SimpleMotion` to enable the same optimizations without making users convolute their code with more glue.

PR #248, opened 2015-12-30
Added sequences of Motion

ccoffey:
This PR enables sequences of Motion and an onCompleted callback which fires after each one completes.

``` javascript
getStyle(){
    return [
      {x: spring(400), y: 0},
      {x: 400, y: spring(400)},
    ];
  },
  onCompleted(){
    console.log('completed!');
  },
  render() {
    return (
        <Motion 
            defaultStyle={{x: 0, y: 0}} 
            style={this.getStyle()} 
            onCompleted={this.onCompleted} >
            {({x, y}) =>
               ...
            }
         </Motion>
```

Please see **demos/demo9-sequence-of-animations** for a full example.

### DIFF
--- a/demos/demo9-sequence-of-animations/Demo.jsx
+++ b/demos/demo9-sequence-of-animations/Demo.jsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import {Motion, spring} from '../../src/react-motion';
+
+const Demo = React.createClass({
+  index: 0,
+  getInitialState() {
+    return {play: false};
+  },
+  handleMouseDown() {
+    this.setState({play: true});
+    document.getElementById('play').disabled = true;
+  },
+  handleTouchStart(e) {
+    e.preventDefault();
+    this.handleMouseDown();
+  },
+  getStyle(){
+    return [
+      {x: spring(this.state.play?  400 : 0), y: 0},
+      {x: 400, y: spring(this.state.play ? 400 : 0)},
+      {x: spring(this.state.play ? 0 : 400), y: 400},
+      {x: 0, y: spring(this.state.play ? 0 : 400)},
+    ];
+  },
+  onCompleted(){
+    this.index ++;
+    if (this.index === 4) {
+      this.setState({play: false});
+    }
+  },
+  render() {
+    return (
+      <div>
+        <button
+          id='play'
+          onMouseDown={this.handleMouseDown}
+          onTouchStart={this.handleTouchStart}>
+          Play
+        </button>
+
+        <Motion defaultStyle={{x: 0, y: 0}} style={this.getStyle()} onCompleted={this.onCompleted}>
+          {({x, y}) =>
+            // children is a callback which should accept the current value of
+            // `style`
+            <div className='demo9'>
+              <div className='demo9-block' style={{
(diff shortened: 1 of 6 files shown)


chenglou:
The completion callback should be per style item, otherwise it's not clear what it means when we have `{a: 10, b: 20}` as `defaultStyle` and `{a: 50, b: 50}` as `style`.

ccoffey:
So that I am clear I understood your comment, in my example above this would mean that onCompleted would fire 4 times?

For the first style in the sequence: {x: spring(400 : 0), y: 0}

1. When y reaches 0 (which is immediately because y is already at its default value)
2. When x reaches 400

For the second style in the sequence: {x: 400, y: spring(400)}

1. When x reaches 400 (which is immediately because this was the final state of x by the end of the first style)
2. When y reaches 400

Alternatively and this probably makes more sense, onCompleted should only fire for springs?

For the first style in the sequence: {x: spring(400 : 0), y: 0}

1. When x reaches 400

For the second style in the sequence: {x: 400, y: spring(400)}

1. When y reaches 400

onCompleted & onStyleItemCompleted
---

I think firing onCompleted per *style item* is useful but I think its a different use case, maybe it make sense to have two callbacks? onStyleItemCompleted and onStyleCompleted?

In my mind, if the user was to provide an array of styles for example

```javascript
[
  {x: spring(400), y: 0},
  {x: 400, y: spring(400)},
]
```

Then they are explicitly stating that they care about the transitions: defaultStyle -> {x: 400, y: 0} -> {x: 400, y: 400}

So I think it makes sense to fire an onStyleCompleted callback between each transition.

What are your thoughts on this?

VasilyShelkov:
I really hope this makes it into a release and look forward to hearing more :D

sdras:
I look forward to this too!

arcanis:
@ccoffey @chenglou Now that `onRest` has been published, do you think it could help this PR moving forward?

uptownhr:
any update on this PR?

lzl124631x:
Looking forward to sequencing in React-Motion. Are @chenglou @appsforartists still maintaining this project? Haven't seen any update for a period of time.

appsforartists:
I'm not a maintainer - I just needed `onRest` for a prototype I was working on, so I sent a patch upstream.  You are welcome to do the same.  😄

sdras:
I talked to @chenglou yesterday about this PR, (and please, Cheng-Lou, let me know if I have anything wrong here I don't mean to speak for you, just want to provide some clarification for people waiting)

He mentioned that this type of sequencing might interfere with spring. Personally, in watching some of his talks, it does seem a little counter-intuitive to his wishes for the direction of the library in terms of true interruptible motion. 

There is another project that he pointed me to, however, that is structurally similar, but allows for sequencing: https://github.com/clari/react-tween

If you're interested in more detailed sequencing with extensive tooling, both GreenSock (here is a piece of an article I wrote showing how to use it in React: https://css-tricks.com/comparison-animation-technologies/#article-header-id-16 or Mo.js, (though this tool is still in beta) https://css-tricks.com/introduction-mo-js/

Anyways, I hope that info is helpful for people waiting

theformer:
这是来自QQ邮箱的假期自动回复邮件。
 

PR #261, opened 2016-01-17
Replace pretty matcher with karma-jasmine-diff-reporter

### LINKED ISSUE #260: prettyMatcher recursion detection is inaccurate
cc @nkbt 
Not really important but if you have `[a, a]` where a is a reference, it'll mistakenly label the second as `[*RECURSION*]`, since we check the cache but never check whether that reference is embedded inside itself.
We can use a json serializer that supports recursion. I've seen at least one.
Filing this as really low-pri, also to remind us to  check if we can use a pre-made reporter now. I refuse to believe there isn't a pre-made reporter that fits our simple needs so far =).

nkbt:
Fixes #260

Before:

After:

I also submitter a PR to override default background color, so we can make it less crazy (optionally): https://github.com/mradionov/karma-jasmine-diff-reporter/pull/8


nkbt:
Updated, so now it looks bit better:

nkbt:
Ping @chenglou ok?

chenglou:
God damn it I keep missing these pull requests!
Thanks a lot.

nkbt:
no problem. I guess pings work =)

PR #269, opened 2016-01-27
Add /lib/.gitignore to fix build error

appsforartists:
If you try to build currently from a fresh clone, you get:

```
rm: lib/*: No such file or directory
```

This fixes that by tracking an empty `lib` in git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ pids
 coverage
 node_modules
 bower_components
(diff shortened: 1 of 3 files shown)


appsforartists:
Tracking lib in git is the first step.  The next thing to do is to make sure that the script doesn't bail if `rm` can't find anything to delete.

PR #270, opened 2016-01-27
Add /lib/.gitignore to fix build error

appsforartists:
If you try to build currently from a fresh clone, you get:

```
rm: lib/*: No such file or directory
```

This fixes that by tracking an empty `lib` in git.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "start": "node server.js",
     "build-demos": "webpack",
     "lint": "eslint --ext .js,.jsx .",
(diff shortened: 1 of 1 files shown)


nkbt:
Why not simply `rm -rf ./lib/*`?

appsforartists:
Because when I looked at the man pages, I couldn't find the option to fail silently.  I'll simplify and push again.

nkbt:
Cool, thanks

chenglou:
Hahaha `-rf`. Totally forgot that worked here.
Sorry about that. I've re-aliased `rm` to OS X `trash` so had none of these problems.

Issue #235, opened 2015-11-02
Tell when animation is finished


nkbt:
Sort of related to #100

It is OK now to check current value if you are using default spring preset. But if I am using something like `wobbly` and animate `height`, I can easily get zero height while animation is far from being finished (then it goes to negative height, but that is a different story).

So if I think that my animation is done when height equals 0 and want to unmount element, that is not actually true and on next RAF component is shown again but with a very little height. And again and again.

![animation-end](https://cloud.githubusercontent.com/assets/175264/10896978/80f220d8-8211-11e5-8bca-e33809e44b75.gif)

I am pretty sure we had this discussion somewhere else, but could not find a separate issue for this.

I would like to see API in this way:

```
<Motion>
  {(style, {isFinished}) => console.log(isFinished)}
</Motion>
```

Passing another object to `render` function will not break current API, and will potentially allow us to pass more fields, not only `isFinished` but exposing some RM internal state (maybe related to #219)

bsansouci:
What's the advantage of this over adding a prop `onEnd` which would call a callback? I'm a little scared of this `isFinished` inside the render function because it sounds like you'll have to mutate some state in the parent component to register that the animation is done. Ultimately you want the parent to know that the animation is done, not necessarily the children.

nkbt:
Callback would work for some cases, but in my case having `isFinished` passed to render itself would work better, since I do not want to manipulate state, based on callbacks and I want only change what is rendered inside function without re-rendering whole component. I thing both should be added. Though this one probably should not be added to the main README (and added to a separate Advanced usage WIKI page)

chenglou:
`end` callback and `isFinished` here would need to be per number btw, not per style

sompylasar:
@chenglou @nkbt The #219 velocities are a better way to go than a single `isFinished` flag. There could be a helper to which you pass the object with velocities so it calculates whether all / any / some of them have reached certain threshold.

nkbt:
@sompylasar I mentioned that issue in my first message =).
That could be an additional thing passed to `render` callback alongside with `isFinished` flag, which is simple and straightforward and does not require any extra logic around it.

sompylasar:
@nkbt, yes, I saw this. Just missed the "alongside" thing, sorry.

yanivtal:
Is anyone working on this?

bsansouci:
@chenglou I don't think anyone's working on it as of right now. PRs are very welcome though :D

[...]

PR #273, opened 2016-01-28
Add Motion.@onRest

appsforartists:
I don't intend for this to land until after I've used it and made sure we're solving the right problem, but I'm opening this now so @chenglou can play with it (and maybe add Flow if there are places I've missed).

See also #235.

### DIFF
--- a/src/Motion.js
+++ b/src/Motion.js
@@ -26,6 +26,7 @@ const Motion = React.createClass({
       PropTypes.object,
     ])).isRequired,
     children: PropTypes.func.isRequired,
+    onRest: PropTypes.func,
   },
 
   getInitialState(): MotionState {
@@ -40,6 +41,7 @@ const Motion = React.createClass({
     };
   },
 
+  wasAnimating: false,
   animationID: (null: ?number),
   prevTime: 0,
   accumulatedTime: 0,
@@ -94,11 +96,18 @@ const Motion = React.createClass({
         propsStyle,
         this.state.currentVelocity,
       )) {
+        if (this.wasAnimating && this.props.onRest) {
+          this.props.onRest();
+        }
+
         // no need to cancel animationID here; shouldn't have any in flight
         this.animationID = null;
+        this.wasAnimating = false;
         this.accumulatedTime = 0;
         return;
       }
+
+      this.wasAnimating = true;
 
       const currentTime = defaultNow();
       const timeDelta = currentTime - this.prevTime;
--- a/test/Motion-test.js
+++ b/test/Motion-test.js
@@ -3,6 +3,8 @@ import {spring} from '../src/react-motion';
 import createMockRaf from './createMockRaf';
 import TestUtils from 'react-addons-test-utils';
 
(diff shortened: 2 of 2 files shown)


chenglou:
Thanks a lot!

appsforartists:
My pleasure!

threepointone:
super!

kumarharsh:
wow! awesome!

leopoldjoy:
@appsforartists @chenglou Can this be used for `<Motion />`, `<StaggeredMotion />`, and `<TransitionMotion />`?

chenglou:
@leopoldjoy currently only implemented for Motion, which is why I waited before the imminent public release, because I wanted it on all 3 components. But after discussing with people I've noticed that they only want it for Motion?

leopoldjoy:
Yes, I agree that it seems like it's only really relevant to `Motion`. I was going to open a pull request for the `README.md` documentation as I found it extremely useful, would this be helpful/relevant?

chenglou:
I'll publish it officially this week or next so don't worry about it.

leopoldjoy:
@chenglou Ok, sounds good! Thanks for your work!

sunflowerdeath:
onRest for TransitionMotion could be also very useful.

haraldox:
@chenglou @appsforartists Great work guys! would be awesome to have that in master :)

AlastairTaft:
Would be really useful to have this for `<StaggeredMotion />`. I'm trying to animate page navigation. Each page is made up of a handful of panels, maybe 3 or 4. When the user navigates between pages I want each panel to staggered-ly spring in. I need to know when the animation is done so that I can remove the old page component.

ollwenjones:
Echo @AlastairTaft except for `TransitionMotion`

appsforartists:
I don't think @chenglou or I use this library anymore, so if you do and you want `@onRest` for the other components, you should open a PR. 😃

ollwenjones:
Is there a new flavor of the month for this kind of thing we should look at instead? 😜

appsforartists:
I still like React Motion.  

I've got an experimental motion library that I use for my work, but it's not documented or published, which unfortunately makes it hard for other folks to use.

ollwenjones:
Heh. How it goes for most of us, seems like. My use-case is a side-panel that squishes a map. Ideally, the map can be told to recenter/repaint after the squishing animation completes, but honestly I'm not sure anyone will notice it _isn't happening_ anytime soon. If it becomes an issue that makes it to the top of my backlog, then you can look for a PR from me. ;)

kumarharsh:
See React-Spring or Popmotion (both are in active development). I haven't tried either yet, but they are on my todo list. 😄

nkbt:
We still use react motion in many places. Though the only component that we work with is the main Motion, so I have no context/interest in modifying any other components.
((tbh I would be quite happy to have only that one available and not add up to the bundle))

ollwenjones:
looks like popmotion/pose is a bit higher level, which is nice.

PR #289, opened 2016-02-24
Update the doc about willLeave

### LINKED ISSUE #288: willLeave docs seems off
Hi,

[The doc for willLeave](https://github.com/chenglou/react-motion#--willleave-stylethatleft-transitionstyle--style)  specifies

```
- willLeave?: (styleThatLeft: TransitionStyle) => ?Style

Optional. Defaults to () => null. The magic sauce property.

styleThatLeft: the e.g. {key: ..., data: ..., value: ...} object from the styles array, identified by key, that was present during a previous render, and that is now absent, thus triggering the call to willLeave.

Return: null to indicate you want the TransitionStyle gone immediately. A Style object to indicate you want to reach transition to the specified value(s) before killing the TransitionStyle.
```

Is the parameter "value" an outdated param? Or is that supposed to be present. Only thing I seem to get is style, where style represents the full spring

for example
```
{ style: width: { stiffness: 120, damping: 14, precision: 0.01, val: 100 }
```
Val represent the wanted value, not the current, and I can't seem to get the current computed val. Before we could have access to val & velocity.

If it's an outdated param, I can make a PR to update that part of the doc! Just let me know

dozoisch:
- updated the value property name to style.
- added information about the style property

- fixes #288

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ Required **function**. Similar to other two components' `children`. Receive back
 ##### - willLeave?: (styleThatLeft: TransitionStyle) => ?Style
 Optional. Defaults to `() => null`. **The magic sauce property**.
 
-- `styleThatLeft`: the e.g. `{key: ..., data: ..., value: ...}` object from the `styles` array, identified by `key`, that was present during a previous render, and that is now absent, thus triggering the call to `willLeave`.
+- `styleThatLeft`: the e.g. `{key: ..., data: ..., style: ...}` object from the `styles` array, identified by `key`, that was present during a previous render, and that is now absent, thus triggering the call to `willLeave`. Note that the style property is exactly what you passed in `styles`, and is not interpolated. For example, if you passed a spring for `x` you will receive an object like `{x: {stiffness, damping, val, precision}}`.
 
 - Return: `null` to indicate you want the `TransitionStyle` gone immediately. A `Style` object to indicate you want to reach transition to the specified value(s) before killing the `TransitionStyle`.
 


README.md
```diff
@@ -258,7 +258,7 @@ Required **function**. Similar to other two components' `children`. Receive back
 ##### - willLeave?: (styleThatLeft: TransitionStyle) => ?Style
 Optional. Defaults to `() => null`. **The magic sauce property**.
 
-- `styleThatLeft`: the e.g. `{key: ..., data: ..., value: ...}` object from the `styles` array, identified by `key`, that was present during a previous render, and that is now absent, thus triggering the call to `willLeave`.
+- `styleThatLeft`: the e.g. `{key: ..., data: ..., style: ...}` object from the `styles` array, identified by `key`, that was present during a previous render, and that is now absent, thus triggering the call to `willLeave`. Note that the style property is exactly what you passed in `styles`, and is not interpolated. For example, if you passed a spring for `x` you will receive an object like `{x: {stiffness, damping, val, precision}}`.
```
chenglou: Humm I intended this object to be opaque. The main use-case is you using it to return a value like so: `{key: ..., style: {x: styleThatLeft.x, y: someChangedValue}}`. Do you need to introspect it?

dozoisch: I actually don't need to introspect it. I was expecting to get the current tween'd value somewhere, which was enabling for early "null" return

dozoisch: @chenglou anything you want me to modify?

chenglou: Sorry, got caught in analysis paralysis and then dropped this. I'm thinking whether I should re-expose velocity and all. I've previously exposed too much and now maybe too little.

chenglou:
Thanks!

PR #295, opened 2016-03-03
Update to support Flow 0.22.0

cdebotton:
Flow no longer globally recognizes internal React types.
Manually import type { Element as ReactElement } to squash Flow errors.
Also ignoring the test folders via flow config, they are causing superfluous warnings.

### DIFF
--- a/src/Motion.js
+++ b/src/Motion.js
@@ -7,6 +7,7 @@ import defaultRaf from 'raf';
 import shouldStopAnimation from './shouldStopAnimation';
 import React, {PropTypes} from 'react';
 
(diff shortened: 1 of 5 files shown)


src/Motion.js
```diff
@@ -7,6 +7,7 @@ import defaultRaf from 'raf';
 import shouldStopAnimation from './shouldStopAnimation';
 import React, {PropTypes} from 'react';
 
+import type { Element as ReactElement } from 'react';
```
chenglou: small nit: can you trim the spaces around the braces please? Thanks =)

PR #298, opened 2016-03-06
Flow update

cdebotton:
Fixed the curl brace spacing as requested in https://github.com/chenglou/react-motion/pull/295


chenglou:
Thanks! Btw in the future, you can just force commit to your previous PR, then force push to override.

PR #305, opened 2016-03-22
Try fixing souporserious/react-motion-ui-pack#41 by checking if unmounting is in progress.

Nimaear:
This fixes the TransitionMotion component not unmounting nicely. I am not 100% this is the right way to do this or why I am having so much trouble reproducing this using react-motion only.


chenglou:
Humm we definitely need to test against this. I'm also not sure this is the most "elegant" way to do it, but that'll do for now. We can revisit this later. Thanks!

PR #309, opened 2016-03-27
in TransitionMotion allow number proptype for "key"

superplussed:
The update makes it so that a key can be set as a number on styles and defaultStyles like so:

`<TransitionMotion
    defaultStyles={[{key: 1, style: {x: 0}}]}
    styles={() => [{key: 1, style: {x: spring(10)}}]}>
</TransitionMotion>`

### DIFF
--- a/src/TransitionMotion.js
+++ b/src/TransitionMotion.js
@@ -185,14 +185,20 @@ type TransitionMotionState = {
 const TransitionMotion = React.createClass({
   propTypes: {
     defaultStyles: PropTypes.arrayOf(PropTypes.shape({
-      key: PropTypes.string.isRequired,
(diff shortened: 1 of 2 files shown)


chenglou:
There's actually a reason why keys are string only; see https://github.com/chenglou/react-motion/commit/c44037dea6aebd87a188eea98499a031e2d40093. I'm hesitating to allow number keys here so I'll keep the PR around. Might reconsider in the future.

cht8687:
We can also add some specific warnings if we only allow string for keys.
I've encountered the same issue here and spent some time to figure out this.

I think this is a good PR.

julienvincent:
Alternatively why not allow both `string` and `integer` props and internally cast the key to a string?

PR #335, opened 2016-06-04
Fix flow for TransitionMotion.unmounting

marcelbeumer:
Since 0.4.4 flowtype complains
```
node_modules/react-motion/src/TransitionMotion.js:508
508:     this.unmounting = true;
              ^^^^^^^^^^ property `unmounting`. Property not found in
186: const TransitionMotion = React.createClass({
                              ^ React component
```


chenglou:
Thanks!

PR #361, opened 2016-08-22
Fix Flow ReactElement errors

bgeihsgt:
In later versions of flow, it requires that generic types have their
type parameters specified. All of the `ReactElement` declarations were
throwing an error saying these needed to be specified.

This change updates `ReactElement` to `React$Element<*>`

### DIFF
--- a/src/Motion.js
+++ b/src/Motion.js
@@ -223,7 +223,7 @@ const Motion = React.createClass({
     }
   },
 
-  render(): ReactElement {
+  render(): React$Element<*> {
     const renderedChildren = this.props.children(this.state.currentStyle);
     return renderedChildren && React.Children.only(renderedChildren);
   },
--- a/src/StaggeredMotion.js
+++ b/src/StaggeredMotion.js
@@ -244,7 +244,7 @@ const StaggeredMotion = React.createClass({
     }
   },
 
-  render(): ReactElement {
+  render(): React$Element<*> {
     const renderedChildren = this.props.children(this.state.currentStyles);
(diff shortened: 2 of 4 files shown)


zhigang1992:
@bgeihsgt seems like it's breaking Eslint. maybe use `ReactElement<*>` instead?

bgeihsgt:
@zhigang1992 Looks like it's a parsing error with brackets. I tried `ReactElement<*>` and got the same errors. Looking around, I think you'll need to upgrade `babel-eslint` to at least version 5.0.0. That requires other upgrades too that I tried for a little bit and then gave up b/c I didn't have time to investigate and fix.

jefflau:
Any update on this?

bgeihsgt:
@jefflau someone needs to upgrade babel-eslint to at least 5.0.0 which requires a cascade of other upgrades. I tried to upgrade for a little bit but didn't have the time to see it through.

AgentME:
I've created a new pull request https://github.com/chenglou/react-motion/pull/371 which extends this one and fixes the issues with it.

bgeihsgt:
Cool, thanks @AgentME!

chenglou:
@AgentME to be clear: if I merge this PR then merge #371, things will be fine?

AgentME:
This pull request contains the commits in #371, so only this one is necessary.

AgentME:
@chenglou Yes, that will work.

chenglou:
Let's try it then! Thanks @bgeihsgt and sorry for the wait!

bgeihsgt:
Cool! Thanks!

PR #362, opened 2016-08-27
Added syntax highlighting to first readme example

amilajack:


nkbt:
Thanks, `jsx` highlighting seems to be ok on github, but unsupported on npm

PR #371, opened 2016-09-21
Fix flow react element errors, upgrade eslint, expose Flow type info

AgentME:
This pull request builds off of https://github.com/chenglou/react-motion/pull/361 and fixes the issues with it.

Currently, adding react-motion as a dependency to a project that uses a recent version of Flow triggers Flow errors because Flow now requires type parameters to `React.Element` (needs to be `React.Element<*>` now), and because react-motion has Flow errors that require a specific value be set for Flow's "suppress_comment" option.

#361 addressed the `React.Element` -> `React.Element<*>` issue, but the pull request was broken for two issues: 1) react-motion still uses Babel 5.x doesn't understand Flow's type parameter syntax, and 2) the version of eslint used didn't understand Flow type parameters either.

The first part was addressed by defining a Flow type alias for `React.Element<*>` in a Flow-comment that Babel wouldn't see. (Updating to Babel 6.x could be a good idea in the future.) The second part was addressed by updating eslint. I also updated the other eslint dependencies like "eslint-config-airbnb", which brought along a lot of new stricter rules. I changed a few parts of the code to pass some of the less-controversial rules, but I disabled most of the new failing rules. It might be worth re-enabling those rules and patching the code to meet them, but I don't see a need to group that into this pull request.

I removed the `//$FlowFixMe` comments so that Flow-using consumers no longer need to set the "suppress_comment" option. A bunch of them were only needed because `rehydrateStyles()` was incorrectly annotated as returning `Array<PlainStyle>` when it really returned `Array<TransitionPlainStyle>`. The other ones I replaced with `:any` casts which don't require configuration to work. I left some comments explaining why some were needed.

I've set the CI system to run Flow (in addition to eslint and the tests) by adding flow-bin as a devDependency, adding a script in package.json which launches it, and made .travis.yml use it.

Finally, I tweaked the build to expose react-motion's type information to Flow users that depend on react-motion. People who use Flow and depend on react-motion won't need to configure anything in order to take advantage of it. Flow will type-check the names of imports from react-motion, type-check the props passed to react-motion's components, etc. I did this by changing the prerelease script in package.json to use [flow-copy-source](https://www.npmjs.com/package/flow-copy-source) to create ".flow" files next to all babel-compiled files in lib/. ([react-draggable-list](https://github.com/StreakYC/react-draggable-list) and [ud](https://github.com/AgentME/ud) are examples of other projects with this same setup.)


chenglou:
@AgentME Merged, thanks! Can you check master still works?

AgentME:
Master now has a few lint errors. Might've been caused by some newer commits not in my branch? I'll put up a pull request to fix it.

AgentME:
@chenglou Fix is in https://github.com/chenglou/react-motion/pull/373. I think the issue had to do with #371 containing #361's commit, but because they were both done in rebase mode git didn't recognize they were the same commit and it created conflicts. #373 fixes it.

twobit:
@chenglou can we get a new build? Would love to pull in this change, thanks @AgentME!

chenglou:
Done! Tell me if there's anything wrong.

PR #397, opened 2016-11-23
More descriptive variable name

therewillbecode:
Changed "delta" to topDeltaY as variable describes the delta of y coordinated from the first list item.


therewillbecode:
Does anyone know why I am getting the following eslint error? 

/home/travis/build/chenglou/react-motion/test/index.js
 1:51  error  Unnecessary escape character: \-  no-useless-escape

nkbt:
Must be updated eslint/airbnb, which started to track this thing. This is why I now always lock down all the deps and ideally use shrinkwrap. Feel free to fix that one bit and will get it in.

PR #401, opened 2016-11-26
Performance optimisation for draggable list demo

therewillbecode:
Issue - Originally on every mouse move event the reinsert function was called to reorder list item positions. 

Fix - Added an if clause to ensure that "order" state items are only reordered when positional changes actually occur in the DOM.


chenglou:
Nice! Thanks =D

Issue #375, opened 2016-09-30
Flow: Property not found


codejunkienick:
After an update I have this issue. I check my code func children is there 

```
220:             <Motion 
                 ^ React element `Motion`
 22:   propTypes: {
       ^^^^^^^^^ property `children`. Property not found in. See: ../node_modules/react-motion/lib/Motion.js.flow:22
220:             <Motion 
                 ^ props of React element `Motion`
```

chenglou:
@AgentME do you know what's going on here?

AgentME:
This is https://github.com/facebook/flow/issues/1964

kellyrmilligan:
any ideas on a workaround? the workarounds in the issue are not working for me

AgentME:
Instead of `<Motion>...</Motion>`, you could use `<Motion children={...} />`. Yeah this issue is a bit annoying.

nsimonson:
To make `<Motion>...</Motion>` work you can make children optional.

So `MotionProps` becomes:
```
export type MotionProps = {
  defaultStyle?: PlainStyle,
  style: Style,
  children?: (interpolatedStyle: PlainStyle) => ReactElement,
  onRest?: () => void,
};
```

cdebotton:
Any progress on merging this? Would love to see this resolved as we're relying heavily on flow and this pollutes the actual flow report.

jariz:
Come on guys just mark it optional, I keep running into this issue with every react motion component I use on every project.
I don't fully understand why flow still hasn't fixed the children bug but I'd rather have it marked as optional than me having to add a `//$FlowFixMe` every time I use a react motion element.

AugustinLF:
This is fixed with Flow 0.53. However, flow detects errors in react-motion, due to how the React types evolved :/

PR #408, opened 2016-12-09
Make children prop type optional

nsimonson:
A bug in flow (https://github.com/facebook/flow/issues/1964) prevents it checking in it's children prop has been set in jsx. You can make flow ignorant to this prop by setting it as optional.

Addresses #375


AugustinLF:
This has been released on flow's side, with the 0.53 version.

chenglou:
Fixed by either #485 or #486 I believe. Thanks!

PR #420, opened 2017-01-13
Unmounted component setState fix

alleycat-at-git:
This code in `TransitionMotion` was a source of bug for us in testing. 

```js
    if (this.unmounting) {
        return;
    }
    // TODO: when config is {a: 10} and dest is {a: 10} do we raf once and
    // call cb? No, otherwise accidental parent rerender causes cb trigger
    this.animationID = defaultRaf((timestamp) => {
      const propStyles = this.props.styles;
      ...
      this.setState({
        currentStyles: newCurrentStyles,
        currentVelocities: newCurrentVelocities,
        lastIdealStyles: newLastIdealStyles,
        lastIdealVelocities: newLastIdealVelocities,
        mergedPropsStyles: newMergedPropsStyles,
      });
      ...
    }
```

The problem is that if execution passes the conditional `if (this.unmounting)`, then executes async `defaultRaf` and after that component unmounts and after that the callback of `defaultRaf` is called, then `setState` will be called on unmounted component.

This fix solves the problem.

### DIFF
--- a/src/TransitionMotion.js
+++ b/src/TransitionMotion.js
@@ -342,12 +342,13 @@ const TransitionMotion = React.createClass({
   },
 
   startAnimationIfNecessary(): void {
-    if (this.unmounting) {
-      return;
-    }
     // TODO: when config is {a: 10} and dest is {a: 10} do we raf once and
     // call cb? No, otherwise accidental parent rerender causes cb trigger
     this.animationID = defaultRaf((timestamp) => {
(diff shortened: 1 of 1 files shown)


chenglou:
Thanks. Wouldn't the right fix be having the check in both locations? This way we bail early instead of allocating another callback.

alleycat-at-git:
@chenglou Done!

PR #430, opened 2017-02-14
Fix broken link

codler:


nkbt:
Wow Flowcheck does something odd


node_modules/flow-copy-source/node_modules/kefir/dist/kefir.js.flow:25
 25:   +value?: (value: V) => void;
       ^ Unexpected token +

PR #439, opened 2017-03-06
Refactored chat heads demo

therewillbecode:
Added controls for spring parameters in order to make the demo even more fun. Also provides users with a convenient way of exploring spring configurations for the staggered motion component.


therewillbecode:
www gifcreator me_ldrl38

demos/demo1-chat-heads/Demo.jsx
```diff
@@ -1,40 +1,189 @@
...
+    ];
+
     return (
+      <div>
+        <div className="header"  ref={element => this.divRef = element}>
+          <div className="parent-container">
+            <div className="dropdown-container">
+              <div className=".mdl-selectfield">
+                <label>Stiffness Behaviour</label>
+                <Dropdown className='Dropdown'
```
nkbt: Why not just `<select>`? What is the benefit of `Dropdown`)? I know that it uses select itself, but there is also heaps of extra logic I struggle to understand need for

Is it possible to make these customizations without creating extra ui components, keeping example maybe a bit less stylish, but more focused around react-motion itself?

therewillbecode: I just thought that adding extra shinyness to the demo would help further the adoption of react-motion. I can try and condense all the additional logic back into the original component if you want?

nkbt: Usually the point of examples is to be more like "bettere" documentarion. Something, people can copy to their project and play with. So it would be better if  each example has minimal amount of stuff around actually useful code. I reckon, nice and pretty thing is great for demo/conference, but hurts newcomers who seek minimal cases they can start working with.

This is my personal opinion of course.

nkbt: I'm still horrible at typing on my phone 📱 😭

therewillbecode: Ok thanks. I'll update the pull request and make sure that the bare minimum logic is there to add the spring customisations.

demos/demo1-chat-heads/Demo.jsx
```diff
@@ -1,59 +1,222 @@
...
+    description: "Linear",
+    value: "linear"
+  },
+  {
+    description: "Inverse Linear",
+    value: "invertedLinear"
+  }
+];
+
+const Demo = React.createClass({
```
Inlesco: If this PR is targeted at a complete refactoring, maybe consider replacing the outdated React.createClass with a ES6 class?

demos/demo1-chat-heads/Demo.jsx
```diff
@@ -1,59 +1,222 @@
...
+  getInitialState() {
```
Inlesco: Consider using ES6 class-fields, so this `getInitialState()` method basically becomes a declaration of `state = { ...initialState }`? See: https://babeljs.io/docs/plugins/transform-class-properties/

demos/demo1-chat-heads/Demo.jsx
```diff
@@ -1,59 +1,222 @@
...
-    window.addEventListener('mousemove', this.handleMouseMove);
-    window.addEventListener('touchmove', this.handleTouchMove);
-  };
+    window.addEventListener("mousemove", this.handleMouseMove);
+    window.addEventListener("touchmove", this.handleTouchMove);
+    this.setState({ headerHeight: this.divRef.clientHeight });
+  },
+
+  shouldComponentUpdate(nextProps, nextState) {
+    if (this.state.x !== nextState.x) {
```
Inlesco: Consider putting all of `if` clauses that evaluate under `true` into a single `if` clause with an `OR` operator, thus, reducing the amount of code and repetition. Thoughts?

demos/demo1-chat-heads/Demo.jsx
```diff
@@ -1,59 +1,222 @@
...
+    this.setState({ stiffnessBehaviour: event.target.value });
+  },
+
+  handleDampingParamChange(event) {
+    this.setState({ dampingBehaviour: event.target.value });
+  },
+
+  getStyles(prevStyles) {
     // `prevStyles` is the interpolated value of the last tick
+    let stiffness = this.state.stiffness;
```
Inlesco: Usage of desctructuring? `let { stiffness, damping, precision } = this.state`

Also, `let { ..., dampingBehaviour, stiffnessBehaviour } = this.state` for simplification.

demos/demo1-chat-heads/Demo.jsx
```diff
@@ -1,59 +1,222 @@
...
+    let damping = this.state.damping;
+    let precision = this.state.precision;
+
     const endValue = prevStyles.map((_, i) => {
+      if (this.state.stiffnessBehaviour === "linear") {
```
Inlesco: I see it's possible to use a ternary operator to simplify the code here. 
For example, you can cover `this.state.stiffnessBehaviour === linear` and `this.state.stiffnessBehaviour === inverseLinear` comparisons like this:

`stiffness = stiffnessBehaviour === linear ?  this.state.stiffness * i : this.state.stiffness / i;`

demos/demo1-chat-heads/Demo.jsx
```diff
@@ -1,59 +1,222 @@
...
+            x: spring(prevStyles[i - 1].x, { stiffness, damping, precision }),
+            y: spring(prevStyles[i - 1].y, { stiffness, damping, precision })
+          };
     });
     return endValue;
-  };
+  },
 
   render() {
+    const selectOptions = options.map(({ value, description }) =>
```
Inlesco: Make `selectOptions`, `stiffnessSelect`, `dampingSelect ` and `precisionInput` class methods, so they're not created as functions on each render (thus, cluttering the GC)?

demos/demo1-chat-heads/Demo.jsx
```diff
@@ -1,59 +1,222 @@
...
+} from "../../src/react-motion";
+import range from "lodash.range";
+import Slider from "./Slider";
 
-export default class Demo extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {x: 250, y: 300};
-  };
+const options = [
```
Inlesco: To reduce the amount of code, maybe it's worthy to make a reusable fn for generating all the `options` available?

Like (you can choose any value separator you need, I picked `_` as it's more readable on longer strings):

`makeOptions = (options) => options.map(option => ({ description: option, value: option.replace(' ', '_').toLowerCase() }))`

So your options become a concise `const options = makeOptions([ 'Constant', 'Linear', 'InverseLinear' ]);`

demos/demo1-chat-heads/Demo.jsx
```diff
@@ -1,59 +1,222 @@
...
-  getStyles = (prevStyles) => {
+  handleDampingChange(event) {
+    this.setState({ damping: event.target.value });
+  },
+
+  handleStiffnessParamChange(event) {
+    this.setState({ stiffnessBehaviour: event.target.value });
+  },
+
+  handleDampingParamChange(event) {
```
Inlesco: I'd consider rewriting this as this (utilizing all the good things that ES6 gives us): 

`handleDampingParamChange = ({ target: { value } }) => this.setState({ stiffnessBehaviour: value });`

Or just:

`handleDampingParamChange = (e) => this.setState({ stiffnessBehaviour: e.target.value });`

demos/demo1-chat-heads/Demo.jsx
therewillbecode: Good idea. Perhaps even better would be to use PureComponent since it does a shallow comparison for you, removing the need to explicitly check if state or props have changed?

demos/demo1-chat-heads/Demo.jsx
therewillbecode: The only thing is that there are three possible options so wouldn't the ternary operator look a bit messy in this case?

therewillbecode:
Thanks so much for the feedback! @Inlesco

demos/demo1-chat-heads/Demo.jsx
Inlesco: @therewillbecode that's a good idea. I'd love to see that implemented.

therewillbecode:
@Inlesco Cheers! It is implemented now.

therewillbecode:
This has been open for a while. Anyone happy to review and merge?

therewillbecode:
This has been open for a while. Anyone happy to review and merge? @Inlesco @nkbt @chenglou

PR #446, opened 2017-04-08
Migrated deprecated React.PropTypes and React.createClass

### LINKED ISSUE #445: Migrate from React.createClass to create-react-class
React just released version 15.5 which includes a deprecation of `React.createClass`. See: https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.createclass

If desired, I can make a pull request for this.

Andarist:
Fixes #445

I think we should migrate to the es6 `class` syntax and stop using `React.createClass`, but that would be another step. For now just created this PR, so people stop having deprecation warnings after migration to React v15.5. 

If you feel the same about `class`, I'll create an issue about it, so the matter won't get forgotten and neglected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,9 @@ Fortunately, `TransitionMotion` has kept `c` around and still passes it into the
 This time, when mapping through the two remaining interpolated styles, you'll produce only two components. `c` is gone for real.
 
 ```jsx
-const Demo = React.createClass({
+import createReactClass from 'create-react-class';
+
+const Demo = createReactClass({
   getInitialState() {
     return {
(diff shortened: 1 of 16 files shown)


Dante-101:
When can the users expect this change in a release?

Calyhre:
Also, bumping `create-react-class` to it's latest version will fix the dependencies, and the CI error.

Andarist:
@Calyhre bumped :)

nkbt:
```
+ react-motion@0.4.8
```

nkbt:
Changes https://github.com/chenglou/react-motion/releases/tag/v0.4.8

PR #457, opened 2017-04-25
Use ES6 classes in ‘Motion.js’

ifndefdeadmau5:
Hi,

`npm install` is broken and it seems previous migration PR missed to fix one file `Motion.js`
This PR is trying to fix this. I tested `npm start` on my local machine and worked well.
Thanks

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
   "author": "chenglou",
   "license": "MIT",
   "dependencies": {
+    "create-react-class": "15.5.2",
     "performance-now": "^0.2.0",
     "prop-types": "^15.5.8",
     "raf": "^3.1.0"
--- a/package.json
+++ b/package.json
@@ -73,7 +73,6 @@
   "author": "chenglou",
   "license": "MIT",
   "dependencies": {
-    "create-react-class": "15.5.2",
     "performance-now": "^0.2.0",
     "prop-types": "^15.5.8",
     "raf": "^3.1.0"
--- a/src/Motion.js
+++ b/src/Motion.js
@@ -40,7 +40,7 @@ export default class Motion extends React.Component {
   }
 
   wasAnimating: boolean = false;
-  animationID: ?number = 0;
+  animationID: ?number = null;
   prevTime: number = 0;
   accumulatedTime: number = 0;
 
--- a/src/Motion.js
+++ b/src/Motion.js
@@ -54,7 +54,7 @@ export default class Motion extends Component {
   wasAnimating: false;
   animationID: ?number = null;
   prevTime = 0;
-  accumulatedTime= 0;
+  accumulatedTime = 0;
   // it's possible that currentStyle's value is stale: if props is immediately
   // changed from 0 to 400 to spring(0) again, the async currentStyle is still
   // at 0 (didn't have time to tick and interpolate even once). If we naively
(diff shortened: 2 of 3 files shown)


nkbt:
👍 
I wonder how our tests did not fail

package.json
```diff
@@ -73,6 +73,7 @@
   "author": "chenglou",
   "license": "MIT",
   "dependencies": {
+    "create-react-class": "15.5.2",
```
nkbt: mmm, probably not intentional

src/Motion.js
```diff
@@ -110,7 +117,7 @@ const Motion = createReactClass({
         return;
       }
 
-      this.wasAnimating = true;
+      this.wasAnimating = false;
```
nkbt: ouch

nkbt: this is probably why tests fail

ifndefdeadmau5: I changed this to avoid flow check error(current error message on travis), but I guess it causes karma error... I'm not farmiliar with Flow and internal logic of this file, Could you give me an idea how to fix this flow error?

src/Motion.js
```diff
@@ -20,8 +19,8 @@ type MotionState = {
   lastIdealVelocity: Velocity,
 };
 
-const Motion = createReactClass({
-  propTypes: {
+export default class Motion extends Component {
```
ifndefdeadmau5: Got it!

src/Motion.js
```diff
@@ -42,22 +49,22 @@ const Motion = createReactClass({
...
       lastIdealVelocity: currentVelocity,
     };
-  },
+  }
 
-  wasAnimating: false,
-  animationID: (null: ?number),
-  prevTime: 0,
-  accumulatedTime: 0,
+  wasAnimating: false;
```
nkbt: These should probably go to constructor.

nkbt: This is why flowtype fails. `wasAnimating` and others are prototype properties, not instance properties. So as soon as you put them all to constructor like `this.wasAnimating = false`, tests will pass (I think so)

ifndefdeadmau5: Hmm I put `this.wasAnimating = false` first, and it gets this error

```javascript
src/Motion.js:120
120:       this.wasAnimating = true;
                ^^^^^^^^^^^^ property `wasAnimating`. Property not found in
 22: export default class Motion extends React.Component {
                          ^^^^^^ Motion
```

nkbt: Need to update types to include it, I reckon.

I am not doing much of Flow and never used it with JS Classes, so cannot help much in this case. Maybe someone can give you better advice.

Or there must be some examples of using Flow with React Classes somewhere

ifndefdeadmau5: Okay! that sounds quite promising, will research on that.

src/Motion.js
```diff
@@ -42,22 +50,21 @@ const Motion = createReactClass({
...
       lastIdealVelocity: currentVelocity,
     };
-  },
+  }
 
-  wasAnimating: false,
-  animationID: (null: ?number),
-  prevTime: 0,
-  accumulatedTime: 0,
+  animationID: ?number = null;
```
nkbt: I guess all instance properties, not only wasAnimating...

ifndefdeadmau5: Yes :) I was just testing with one property to see if different result comes out compare to my local test

src/Types.js
```diff
@@ -37,6 +37,10 @@ export type MotionProps = {
   style: Style,
   children: (interpolatedStyle: PlainStyle) => ReactElement,
   onRest?: () => void,
+  wasAnimating?: boolean,
```
nkbt: These properties are not MotionProps, they are instance properties of Motion

ifndefdeadmau5: Oh, so sounds like I need to declare this
```javascript
type Motion = {
  wasAnimating?: boolean,
  animationID?: number,
  prevTime: number,
  accumulatedTime: number,
}
```
in the `Motion.js` file itself?

nkbt:
hm, here is info on how to use instance props.

https://flow.org/en/docs/types/classes/

Looks like it was initially right. Probably needed `:boolean` type for wasAnimating or something like that...

ifndefdeadmau5:
Yeah, that links makes lot of sense! Thank you so much. Learned bit of flow types today.

nkbt:
Me too. I am going to use flow for react components too. So far was using it for only library code

ifndefdeadmau5:
Me either.Did not notice the benefit of Flow, just used proptypes in company project so far

ifndefdeadmau5:
Thank you for supporting this whole journey with me @nkbt. This PR was my first open source contribution

nkbt:
🎉

MichaelDeBoey:
Test are still broken :/
Will be fixed in #462 :-)

PR #460, opened 2017-04-25
Split tests in Travis and use Node 6

nkbt:
Would get better logs in Travis. And Node 4 is a bit dated...


nkbt:
Oh, wow, travis had react-create-class cached so tests passed, though we removed this dependency =(... Need to update tests

PR #462, opened 2017-04-25
Update deprecated 'create-react-class' dependencies in tests to ES6 classes

### LINKED ISSUE #461: Update tests to not use create-react-class
Discovered that as part of https://github.com/chenglou/react-motion/pull/460

Caching `node_modules`  did us a bad service

MichaelDeBoey:
Fixes #461

CC: @nkbt

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -11,7 +11,7 @@
     "build/react-motion.map"
   ],
   "dependencies": {
-    "react": ">=0.13.2 || ^0.14 || ^15.0.0"
+    "react": "^0.14.9 || ^15.3.0"
   },
   "keywords": [
     "react",
--- a/test/Motion-test.js
+++ b/test/Motion-test.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import createReactClass from 'create-react-class';
 import {spring} from '../src/react-motion';
 import createMockRaf from './createMockRaf';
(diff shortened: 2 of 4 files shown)


nkbt:
Awesome!

MichaelDeBoey:
@nkbt just have to fix the linting errors now :P

MichaelDeBoey:
@nkbt Fixed :-)

bower.json
```diff
@@ -11,7 +11,7 @@
     "build/react-motion.map"
   ],
   "dependencies": {
-    "react": ">=0.13.2 || ^0.14 || ^15.0.0"
+    "react": "^0.14.9 || ^15.3.0"
```
nkbt: oh, thanks for that

MichaelDeBoey: You're welcome 😃

test/Motion-test.js
```diff
@@ -1,5 +1,5 @@
+/* eslint-disable class-methods-use-this */
```
nkbt: "fixed" =)

Apparently we can make those methods static according to http://eslint.org/docs/rules/class-methods-use-this

But that is OK for the moment. I'd rather get tests working now.

MichaelDeBoey: Yes indeed, but render isn't supposed to be static so that's why I did it like this

Issue #456, opened 2017-04-24
React 16 support?


kesha-antonov:
Hello!

RN 0.43 require to use react-16.

It'll be great to update this lib to use React 16.
Can we do that?

nkbt:
Should be compatible now. The only annoyance is not matching peer
dependency but that is not a blocker.

It makes sense to add react 16 to peer deps tho. Together with es6 classes
and publish major release

-- 
Nik Butenko
nik@butenko.me
http://github.com/nkbt

nkbt:
I checked and it looks like we still can maintain older versions of React as peerDeps, so it should be safe to add v16 to the peers without major release.

kesha-antonov:
@nkbt Nice! Thanks!

MichaelDeBoey:
`React v16.0` isn't released yet, but everything that works in `v15.5.0`, [will keep working](https://facebook.github.io/react/blog/#new-deprecation-warnings) in `v16.0`, so this can be closed I think :-)

CC: @nkbt

sammkj:
@MichaelDeBoey what about `<TransitionMotion />`? Do we still need a `<div />` to wrap child components?

bezreyhan:
Do we know when the support for React 16 will be released. I'm getting the following error after upgrading to React 16:

`Cannot read property 'objectOf' of undefined`

The error is happening in `Motion.js:44` because of PropTypes:

```
defaultStyle: _react.PropTypes.objectOf(_react.PropTypes.number)
```

nkbt:
Motion with `prop-types` was release quite a long time ago.

bezreyhan:
I realized the problem is coming from swagger-ui which is using an older version of react-collapse v2.3.1

Sorry about that.

nikglavin:
Are you able to tag a version from Master that "supports" React v16?

`"react-motion": "0.5.1",`
`error react-motion#react@^0.14.9 || ^15.3.0 doesn't satisfy found match of react@16.0.0`

This has been resolved on the master branch.

nkbt:
I've published 0.5.2 with react 16 peerDependency, give it a shot!

PR #463, opened 2017-04-25
Use 'react-dom/test-utils' package instead of 'react-addons-test-utils'

### LINKED ISSUE #459: Use 'react-dom/test-utils' package instead of 'react-addons-test-utils'
`react-addons-test-utils` has been deprecated and has been put in the `react-dom` package since [React v15.5.0](https://facebook.github.io/react/blog/#react-test-utils).

MichaelDeBoey:
Closes #459

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -11,7 +11,7 @@
     "build/react-motion.map"
   ],
   "dependencies": {
-    "react": "^15.5.0"
+    "react": "^0.14.9 || ^15.3.0"
   },
   "keywords": [
     "react",
--- a/bower.json
+++ b/bower.json
@@ -11,7 +11,7 @@
     "build/react-motion.map"
   ],
   "dependencies": {
-    "react": "^0.14.9 || ^15.3.0"
+    "react": "^15.5.0"
   },
   "keywords": [
     "react",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "A spring that solves your animation problems.",
   "main": "lib/react-motion.js",
   "peerDependencies": {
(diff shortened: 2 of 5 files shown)


bower.json
```diff
@@ -11,7 +11,7 @@
     "build/react-motion.map"
   ],
   "dependencies": {
-    "react": "^0.14.9 || ^15.3.0"
+    "react": "^15.5.0"
```
nkbt: Since it is only related to tests, maybe worth keeping peer deps as is, but only change devDeps?

MichaelDeBoey: OK  no problem :-)

nkbt: Thanks! 15.5 would be probably too harsh, I doubt that many projects got react update. Actually our projects are still on previous 15.x release.

MichaelDeBoey:
This also fixes #456 I think?

nkbt:
#456 requires react@16 in peer deps to be closed. Will need to check that it actually works fine with 16 with no issues.

MichaelDeBoey:
@nkbt `React v16.0` isn't released yet, but everything that works in `v15.5.0`, [will keep working](https://facebook.github.io/react/blog/#new-deprecation-warnings) in `v16.0`

nkbt:
alpha is there

```
23:56 $ npm info react dist-tags

{ latest: '15.5.4',
  '0.10.0-rc1': '0.10.0-rc1',
  '0.11.0-rc1': '0.11.0-rc1',
  next: '16.0.0-alpha.11',
  dev: '15.5.0-rc.2',
  '0.14-stable': '0.14.9' }
```

PR #466, opened 2017-05-06
Linebreak rule is missing in .eslintrc which causes webpack to fail on Windows...

roz0n:
... simply adding the rule to the eslint config solves the problem.


roz0n:
My apologies, I lacked a full understanding of linebreaks. I'll close this now. Thanks for the clarification!

nkbt:
No worries! 🤝

PR #485, opened 2017-08-23
Upgrade typings to flow 0.53.1

AugustinLF:
I simply ran the `flow-upgrade` command, and fixed the `ReactElement` types. I didn't change anything else, perhaps you'd like me to replace the `import  * as React` (which is what flow-upgrade does) by more granular imports?

### DIFF
--- a/src/Motion.js
+++ b/src/Motion.js
@@ -5,10 +5,10 @@ import stepper from './stepper';
 import defaultNow from 'performance-now';
 import defaultRaf from 'raf';
 import shouldStopAnimation from './shouldStopAnimation';
-import React from 'react';
+import * as React from 'react';
 import PropTypes from 'prop-types';
 
-import type {ReactElement, PlainStyle, Style, Velocity, MotionProps} from './Types';
+import type {PlainStyle, Style, Velocity, MotionProps} from './Types';
 
 const msPerFrame = 1000 / 60;
 
@@ -230,7 +230,7 @@ export default class Motion extends React.Component<MotionProps, MotionState> {
     }
   }
 
-  render(): ReactElement {
+  render(): React.Node {
     const renderedChildren = this.props.children(this.state.currentStyle);
     return renderedChildren && React.Children.only(renderedChildren);
   }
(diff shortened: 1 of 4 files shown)


chenglou:
Oops, sorry I didn't see this one and merged #486 first...
Can you check if that PR is suitable?

AugustinLF:
The PR was correct, but didn't take into account the `React.Node` type, which I believe is more suited for the return value of the children function.

chenglou:
Sounds fair. Thanks!

chenglou:
Wait, no, `React.Children.only` returns a `ReactElement`: https://github.com/facebook/react/blob/cd52f7ed80e23b569634d34e42f8a6ae5948444b/src/isomorphic/children/onlyChild.js#L28

AugustinLF:
Yep, you're right, my bad, didn't consider that `React.Node` is a `React.ChildrenArray`, so a children function with my definition could return `[<div />, <div />]`, which is not compatible.

Sorry for the trouble!

chenglou:
No worries, thanks for the PR still =)

dapetcu21:
The problem with the `ReactElement` typing is that it doesn't allow `null`, falsy or string values, which are still valid return values of `render()`. For example, in my code I have:

```
<Motion style={{ opacity: spring(someCondition ? 0 : 1) }}>
  {({ opacity }) => opacity !== 0 && (
    <SomeComponent />
  )}
</Motion>
```

Which fails the type check because it can return `false`.

AugustinLF:
@dapetcu21 I created the PR (#497) to fix this, could you test your use cases against my fork (the branch is `flow/onlyChildren`)?

dapetcu21:
Cool! I'll do that, but I can't right now. I'll check it out on Wednesday when I get back to work

AugustinLF:
@dapetcu21 any update on that?

dapetcu21:
Oh. Sorry. I forgot about this. I'll add a reminder for tomorrow.

dapetcu21:
Yup. I can confirm #497 solves my issue. Sorry for taking this long. Things came up and I forgot about this issue.

PR #487, opened 2017-08-30
Revert "Upgrade typings to flow 0.53.1"

chenglou:
Reverts chenglou/react-motion#485

onlyChild returns a ReactElement


RobinClowers:
What was the problem? We need flow 53 compatibility, can I help?

PR #490, opened 2017-09-19
Multiple draggable list

xcv58:
Add a multiple draggable list demo


xcv58:
The gif:

xcv58:
There is a demo: https://demo9-multiple-draggable-list-incpdhbwdg.now.sh/

sompylasar:
@xcv58 Looks great! There's still room for improvement on mobile, see a couple screen recordings from Chrome on Android: https://youtu.be/E7GcCtbbOQU https://youtu.be/BSI3_waHh2w

xcv58:
@sompylasar Thank you! Could you double whether the official demo has the same issue? http://chenglou.github.io/react-motion/demos/demo8-draggable-list/

If yes, we can create another issue to fix it. Otherwise, I have to borrow an Android device to reproduce the issue.

sompylasar:
@xcv58 Yes, the [official demo](http://chenglou.github.io/react-motion/demos/demo8-draggable-list/) has the same "pull to refresh" on Chrome Android when dragging down. I'd like to note that for some reason the original one is jumpy when "pull to refresh" is activated (when start by dragging down), but this multiple one is smooth in both directions.

https://docs.google.com/document/d/12Ay4s3NWake8Qd6xQeGiYimGJ_gCe0UMDZKwP9Ni4m8/edit#heading=h.10i1fpz2uwnh  https://stackoverflow.com/a/29313685/1346510

> #### Preventing the pull-to-refresh effect
>
> The default action of the pull-to-refresh effect can be effectively prevented by doing any of the following :
> 1. Applying “touch-action: none” to touch-targeted elements, where appropriate, disabling default actions (including pull-to-refresh) of the touch sequence.
> 2. Applying “overflow-y: hidden” to the body element, using a div for scrollable content if necessary[3].
> 3. preventDefault’ing some portion of the touch sequence, including any of the following (in order of most disruptive to least disruptive):
>     - a. The entire touch stream (not ideal).
>     - b. All top overscrolling[4] touchmoves.
>     - c. The first top overscrolling touchmove.
>     - d. The first top overscrolling touchmove only when 1) the initial touchstart occurred when the page y scroll offset was zero and 2) the touchmove would induce top overscroll.
> 4. Disabling the effect locally via chrome://flags (disable-pull-to-refresh-effect).
> 
> Note that the pull-to-refresh effect will never activate if any scrolling occurs before the overscrolling motion, e.g., if the user first scrolls down, then back up, or if the page scroll offset is 0 but has a scrolling div that the user scrolls up before overscrolling.

xcv58:
@sompylasar Could you create another issue for this so that we can fix all Android Chrome pull problem together?

sompylasar:
@xcv58 https://github.com/chenglou/react-motion/issues/493

xcv58:
@sompylasar thank you!

xcv58:
I finally get an Android phone. Will try to fix this soon.

PR #492, opened 2017-09-21
Add check for undefined values in style object

mikkom:
In our mocha unit tests react-motion caused errors such as:

```
  1)  Uncaught error outside test suite:
     Uncaught TypeError: Cannot read property 'val' of undefined
      at Object.shouldStopAnimation [as default] (node_modules/react-motion/lib/shouldStopAnimation.js:20:78)
      at Object.callback (node_modules/react-motion/lib/Motion.js:117:45)
      at Timeout._onTimeout (node_modules/raf/index.js:35:21)
```

This was caused by cases where the `style` object had undefined values, for example `{ height: undefined }`. Adding an extra guard to shouldStopAnimation seems to fix the issue.

### DIFF
--- a/src/shouldStopAnimation.js
+++ b/src/shouldStopAnimation.js
@@ -9,7 +9,7 @@ export default function shouldStopAnimation(
   currentVelocity: Velocity,
 ): boolean {
   for (let key in style) {
-    if (!Object.prototype.hasOwnProperty.call(style, key)) {
+    if (!Object.prototype.hasOwnProperty.call(style, key) || typeof style[key] === 'undefined') {
       continue;
     }
 


chenglou:
Thanks! Though it's been a while and I forgot whether this was intentional (ugh javascript...). Why are there `undefined` values in the first place?

nkbt:
I feel like failing on any non-number value is actually a good thing

Skovy:
Will this PR be accepted? I'm seeing this same issue.

austinbiggs:
I'm just here to bump this PR 😸 cc @chenglou

theformer:
这是来自QQ邮箱的假期自动回复邮件。
 

PR #498, opened 2017-10-02
React 16 for demos

nkbt:
react-hot-reload 1.x does not support React 16 (uses react/lib/*).

As far as I tried react-hot-reload 3.x it was not working as expected, so at this point I will remove hot-reload entirely and we can re-iterate later.

Used yarn to install deps and added yarn.lock. Updated travis config to use yarn too (and node 8)


chenglou:
Oh well ¯\_(ツ)_/¯

Thanks, merged!

chenglou:
Actually the build is failing?

nkbt:
Yeah, there is some issue with phantomjs. I need to dig a bit more

nkbt:
I've reverted travis back to node6 and npm for now. Don't have much time these days =(. When I'm back from holidays, will update all deps and make sure everything builds properly (and fast).

nkbt:
Oh dear what's wrong this time?

```
02 10 2017 23:24:32.292:WARN [launcher]: Can not load "PhantomJS", it is not registered!
  Perhaps you are missing some plugin?
```

PR #510, opened 2017-12-11
Adding unmount logic for motion

jnarwold:
Added mounting logic to the motion library to help avoid raf issues when removing/adding motion from a react application

### DIFF
--- a/src/Motion.js
+++ b/src/Motion.js
@@ -37,6 +37,7 @@ export default class Motion extends React.Component<MotionProps, MotionState> {
   }
 
   wasAnimating: boolean = false;
+  unmounting: boolean = false;
   animationID: ?number = null;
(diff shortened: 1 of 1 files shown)


nkbt:
Any more details on why it is necessary?

There should not be any real logic since `defaultRaf.cancel(this.animationID)` exists, unless app code performs some side-effects

jnarwold:
This may have been a side effect of our particular use case. When we transitioned between screen sizes we had to remove/re-add the motion component based on screen size. Often we ended up with the following error loop as a result of that transition. It appeared that setState was getting called during the transition on an unmounted component. 

image

theformer:
这是来自QQ邮箱的假期自动回复邮件。
 

PR #511, opened 2017-12-12
Added module entry point + udpated bunch of deps

Andarist:
I've upgraded bunch of stuff, notable changes:
- added `module` entry point - can be tree-shaken and scope hoisted by webpack
- using babel@7-beta
- umd build built by rollup (smaller bundle size)
- lots of deps updated

Source code changes:
- lint related, no functionality changes

Breaking changes:
- for commonjs users: if one is reaching for internal files (instead of importing from the entry point), he/she will need to access default export with `default` propert like this: `const Motion = require('react-motion/lib/Motion').default`

TODO:
- [ ] fix tests

PhantomJS is crashing for some reason (i suspect memory leak in phantom), it is unmaintained at the moment, I'd like to change it to headless Chrome or similar, WDYT?

I'd also advise to create flat bundles for commonjs and es targets (bundle them with `rollup` to a single file, instead of transpiling file by file with `babel`). WDYT?


nkbt:
Looks good.
It is possible to avoid breaking changes and this `default` thing. With either babel transform or manually

Andarist:
@nkbt 
fixed by using https://github.com/59naga/babel-plugin-add-module-exports

WDYT about ditching phantomjs?

nkbt:
I reckon phantomjs is deprecated, isn't it? Don't mind really. Would rather have proper end-to-end tests with selenium instead.

This sounds like an out-of-scope for current PR tho

Andarist:
Ok, gonna work on migrating existing tests to some other headless browser (probably Chrome) when I get time.

Andarist:
Ok, gonna work on migrating existing tests to some other headless browser (probably Chrome) when I get time.

Andarist:
@nkbt Ok, I've finished the transition to headless Chrome, everything seems to work fine :)

Andarist:
@nkbt friendly 🏓 😃

Andarist:
@nkbt friendly 🏓 😃 

Sorry for pinging again, I don't expect you jumping on it right away. I understand that maintaining OSS is a time-consuming burden and I don't expect you working on this, at the same time I know how easy is to miss some notification when you are involved in many projects. Cheers!

Andarist:
@nkbt friendly 🏓

nkbt:
I guess we dragged this PR for so long that webpack released v4 already =)

PS: I am quite out of OSS at this point (at least beyond what I can do during my working hours).  Classic burn out at some point in past...

Andarist:
> I guess we dragged this PR for so long that webpack released v4 already =)

That's true, although this PR is especially relevant because of it - it aims to get most out of webpack's tree-shaking etc. (currently it's not possible because of commonjs format)

> PS: I am quite out of OSS at this point (at least beyond what I can do during my working hours). Classic burn out in the past...

No worries, totally understandable. Not sure if u can commit any time resources for this one, I could do it if you grant me write access and some way to contact u about doing a release. Certainly don't want to commit to maintain this repo, already having too many projects I need to take care of. I can commit to maintaining a tooling around this one though.

nkbt:
I reckon it's only @chenglou who can give permissions

PR #518, opened 2018-01-22
Fixed memory leak

stryju:
This diff achieves two things:
1. applies solution from #420 to `Motion` and `StaggeredMotion`
2. prevents two `rAF`s to be triggered at the same time

# Why?
In some rare cases **ReactFiber** can execute `cWRP` right after `setState` call, which would result in two `rAF`s to be triggered (2nd point addresses this) and this combined with `setState` being triggered after unmounting the component (#420 addressed that, but only for `TransitionMotion`), the `rAF` would never stop, as it would work on stale state, which can result in OOM for low-memory and long-running devices.

### DIFF
--- a/src/Motion.js
+++ b/src/Motion.js
@@ -36,6 +36,7 @@ export default class Motion extends React.Component<MotionProps, MotionState> {
     this.state = this.defaultState();
   }
 
+  unmounting: boolean = false;
   wasAnimating: boolean = false;
   animationID: ?number = null;
   prevTime: number = 0;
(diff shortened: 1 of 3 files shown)


stryju:
Alternative solution would be to re-order the end of `startAnimationIfNecessary` function:

```
      // ...
      this.animationID = null;
      // the amount we're looped over above
      this.accumulatedTime -= framesToCatchUp * msPerFrame;

      this.unreadPropStyles = null;

      this.startAnimationIfNecessary();

      // setState must be called at the end because ReactFiber can invoke
      // cWRP when processing setState
      this.setState({
        currentStyles: newCurrentStyles,
        currentVelocities: newCurrentVelocities,
        lastIdealStyles: newLastIdealStyles,
        lastIdealVelocities: newLastIdealVelocities,
      });
```

chenglou:
Thanks!

PR #534, opened 2018-03-31
Replace deprecated componentWillReceiveProps with componentDidUpdate

alexeyraspopov:
Context: [React v16.3 Component Lifecycle Changes](https://reactjs.org/blog/2018/03/29/react-v-16-3.html#component-lifecycle-changes).

Method `componentWillReceiveProps()` becomes deprecated since it doesn't guarantee the fact component _will_ receive props. In addition, it doesn't seem to be recommended to do any side-effects inside this method.

I attempted to switch to `componentDidMount()` since it seems to be the correct place for such updates and side-effects. I needed to add props check since the method is also called during state updates.

Since `componentDidMount()` is called a moment after `componentWillReceiveProps()`, another render happens between them with current (unchanged) state. This forced me to update a couple of tests with a component's state before actual animation restart. As for the user, it doesn't make any visible difference, since animation starts just a bit (a single event loop spin?) later.

I wanted to apply the same changes to `TransitionMotion` and it mostly worked. There were a couple of tests that I failed to understand and so update accordingly. At this moment, I'd like to get some feedback on the introduced changes and possibly some help moving forward. Thanks!

cc @chenglou @nkbt

### DIFF
--- a/test/Motion-test.js
+++ b/test/Motion-test.js
@@ -539,6 +539,7 @@ describe('Motion', () => {
     mockRaf.step(2);
     expect(count).toEqual([
       {a: 0},
+      {a: 0},
       {a: 0}, // this new 0 comes from owner update, causing Motion to re-render
       {a: 400},
       {a: 385.8333333333333},
@@ -548,9 +549,9 @@ describe('Motion', () => {
       {a: 350.02047519790233},
     ]);
     mockRaf.step(999);
-    expect(count.length).toBe(85);
+    expect(count.length).toBe(86);
     setState({a: spring(400)});
     // make sure we're still updating children even if there's nothing to interp
-    expect(count.length).toBe(86);
+    expect(count.length).toBe(87);
   });
 });
--- a/src/Motion.js
+++ b/src/Motion.js
@@ -5,6 +5,7 @@ import stepper from './stepper';
 import defaultNow from 'performance-now';
 import defaultRaf from 'raf';
 import shouldStopAnimation from './shouldStopAnimation';
+import arePropsEqual from './arePropsEqual';
 import React from 'react';
 import PropTypes from 'prop-types';
 
@@ -210,13 +211,17 @@ export default class Motion extends React.Component<MotionProps, MotionState> {
     this.startAnimationIfNecessary();
   }
 
-  componentWillReceiveProps(props: MotionProps) {
+  componentDidUpdate(prevProps: MotionProps) {
+    if (arePropsEqual(prevProps, this.props)) {
+      return;
+    }
+
     if (this.unreadPropStyle != null) {
       // previous props haven't had the chance to be set yet; set them here
       this.clearUnreadPropStyle(this.unreadPropStyle);
     }
 
-    this.unreadPropStyle = props.style;
+    this.unreadPropStyle = this.props.style;
     if (this.animationID == null) {
       this.prevTime = defaultNow();
       this.startAnimationIfNecessary();
--- a/src/StaggeredMotion.js
+++ b/src/StaggeredMotion.js
@@ -5,6 +5,7 @@ import stepper from './stepper';
 import defaultNow from 'performance-now';
 import defaultRaf from 'raf';
 import shouldStopAnimation from './shouldStopAnimation';
+import arePropsEqual from './arePropsEqual';
 import React from 'react';
 import PropTypes from 'prop-types';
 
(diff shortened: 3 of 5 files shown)


alexeyraspopov:
@chenglou, @nkbt, any chance to get this reviewed? Should I provide some additional rationale?

nkbt:
component did update happens *after* component was updated, component will receive props - *before*. I believe this change may cause a lot of issue in downstream libraries and project and there is no realistic way to test this out. There was `getDerivedStateFromProps` added, but it is static... 

This change would probably require major version bump for this reason.

nkbt:
I haven't started doing any React17 upgrades on any projects I am currently working on and not sure when I will be able to do it. So I cannot properly test such changes in real apps at this point unfortunately.

shalanah:
I think replacing `componentWillReceiveProps` would be a great idea. I'm using react 16.4.1 and this is creating a bug for me when pairing another repo (react-custom-scrollbars wrapping around it). I can confirm the error is with `react-motion` and `componentWillReceiveProps`. I get an error that says I cannot setState on a component that has not yet mounted. And when I comment out `this.startAnimationIfNecessary()` in `StaggeredMotion` in `componentWillReceiveProps` the error disappears.

nkbt (approved):
LGTM

Not sure how much it is going to break downstream though

JanKanty:
Can we expect to be merged soon? :D

JanKanty:
So maybe it's goo idea to for now move to using `UNSAFE` prefix, so without any errors / warnings you can test it more?

alexeyraspopov:
Switching to `UNSAFE` methods is still a breaking change since react-motion currently supports React v0.14 and v15.x. And this breaking change is quite difficult to justify, I'd say.

gnapse:
Hi, any updates on this?

mikestecker:
Any updates on this?

test/Motion-test.js
```diff
@@ -548,9 +549,9 @@ describe('Motion', () => {
       {a: 350.02047519790233},
     ]);
     mockRaf.step(999);
-    expect(count.length).toBe(85);
+    expect(count.length).toBe(86);
```
Justin2997: Why changing this ?

Justin2997 (commented):
Seems to work, We should merge it :)

japalo:
Sorry for repeating what's being asked above, but are there any updated on this?

andrewjrhill:
This really needs priority... give us back our consoles please. Devs shouldn't have to disable warnings in their consoles for an optimal debugging experience. Super annoying.

mlacif1:
We need this to merged quite asap please. Thank you.

chenglou:
@andrewjrhill this repo has been around for more than 4 years now. I'm personally not sure what kind of bugs this'll introduce if we merge and do a release. There are plenty of race conditions that can be had here. @alexeyraspopov how confident are you in the PR?

nkbt:
Yeah I am very much not sure this won’t introduce issues. I’ve seen quite a few problems caused by this sort of change in other components and fixing them was not quite as trivial in many cases unfortunately, as behaviour actually changes significantly

test/Motion-test.js
nkbt: Because component is going to have one more update I assume due to switching to componentDidUpdate

andrewjrhill:
In the case of high-risk changes we could potentially release a "beta" branch with intent to merge with `master` as a new major version at an allocated future date (~3 months?). We can then open an ongoing issue asking people to test this (`@next`?) release and open PR's against it if they discover issues.

theopak:
👋 hey folks! nice idea here, and to add to the discussion i thought it might be nice to consider shipping an `UNSAFE_` lifecycle rename first as a low risk step towards an eventual refactor like this one

and opinions? discussion welcome: #575

japalo:
@theopak as a quick fix, I agree. In a long term perspective I think another fix needs to be implemented. Had I been able to suggest more in-depth solution I would, but sadly I don't have the time right now.

sousajunior:
Was this PR abandoned? Any solution to this?

morsmodr:
> React ensures that any setState calls that happen during componentDidMount and componentDidUpdate are flushed before the user sees the updated UI. In general, it is better to avoid cascading updates like this, but in some cases they are necessary (for example, if you need to position a tooltip after measuring the rendered DOM element).

Not sure whether this snippet from https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#invoking-external-callbacks helps with the concerns regarding this merge

Amirault:
Waiting for the merge and then the release 🙏 🙏 🙏

scotthwsnyder:
Any updates on this?

lazyTai:
你发的邮件，小弟阿泰会尽快查收，么么哒~~~

theformer:
这是来自QQ邮箱的假期自动回复邮件。
 

PR #545, opened 2018-07-04
Upgrade stack

TrySound:
I wanted to upgrade only babel but everything else was broken so I
upgraded babel, webpack, eslint and karma.

Disabled a lot of eslint rules to not change existing code. Maybe it's
better to migrate to prettier later.

This changes will unblock me from providing smaller output.


TrySound:
We need to rerun travis

TrySound:
Friendly ping @chenglou

chenglou:
Thanks for the ping. To be clear, is there a user-facing change here? Or just for us devs?

TrySound:
@chenglou This one only for devs.

TrySound:
Bundle is slightly changed with new versions but I don't think there's anything significant.

TrySound:
@chenglou Tests are passed.

chenglou:
If there's no user-facing impact then sure. Thanks!

PR #546, opened 2018-07-05
Bundle umd with rollup

TrySound:
Rollup produces smaller and readable output which do not require source
maps.

- removed sourcemaps from umd bundle
- bundle umd with rollup
- added production umd version

Previously minified development bundle had 31750b. The new one 24575b


TrySound:
Ready for review. User facing change are removed sourcemaps, smaller umd size + production umd.
/cc @chenglou

chenglou:
Thanks. Is there a way to remove webpack? I guess not since we need it for the demos and watcher?

TrySound:
Yes, let's keep it for now.

chenglou:
Ok. Fwiw I'm more or less blindly merging right now. As long as the demos still work and that we don't break userland (react-motion has been stable for years), I'm good with the cleanups.

Thanks!

PR #547, opened 2018-08-02
:arrow_down: upgrade babel's  configuration file so that remove babel…

strongcode9527:
[Removing Babel's Stage Presets](https://babeljs.io/blog/2018/07/27/removing-babels-stage-presets)

Upgrade the Babel configuration file to match the Babel update


.babelrc
```diff
@@ -1,10 +1,15 @@
...
-    "@babel/stage-3",
     "@babel/flow",
     "@babel/react"
   ],
+  "plugins": [
+    "@babel/plugin-syntax-dynamic-import",
+    "@babel/plugin-syntax-import-meta",
+    ["@babel/plugin-proposal-class-properties", { "loose": false }],
+    "@babel/plugin-proposal-json-strings"
+  ],
```
TrySound: Does anything except class properties is used here?

strongcode9527: yes, and it works well

strongcode9527: [A description of the official document of stage-3](https://github.com/babel/babel/tree/master/packages/babel-preset-stage-3#babelpreset-stage-3)

TrySound: I know about stage-3. I mean does this package contain json strings, dynamic imports or import meta? If not, there's not need for these plugins.

strongcode9527: Thank you for your reply, I understand, and have submitted amendments.

TrySound:
Try also to run `yarn build` and commit size snapshot. We'll see how the size is changed with class properties in loose mode.

strongcode9527:
I've tried it before submitting, and after doing it once, after comparing it with the current version, it turns out that the values are the same as before.

package.json
```diff
@@ -43,7 +43,7 @@
     "react-codemirror": ">=0.1.2",
     "react-dom": ">=15.5.0",
     "rollup": "^0.62.0",
-    "rollup-plugin-babel": "^4.0.0-beta.7",
+    "rollup-plugin-babel": "^4.0.0-beta.2",
```
nkbt: what is the reason for these "downgrades" (+babel-loader)
They both resolved to beta.7 in yarn.lock, but still

strongcode9527: According to the official upgrade command "npx babel-upgrade",  the files will be modified automatically.So I didn't modify these configurations myself.

[Official documents](https://babeljs.io/blog/2018/07/27/removing-babels-stage-presets#migrating)

nkbt: It does not mean it is perfect and we do not need to check if all these automatic changes are actually necessary 🤷‍♂️

TrySound: I upgraded by hands in https://github.com/chenglou/react-motion/pull/548

strongcode9527: ok, thank you for reply, i learn a lot.

PR #548, opened 2018-08-06
Upgrade babel and build esm and cjs bundles

TrySound:
In this diff I suggest to use bundles instead of transpiled with just
babel files. There are a few reasons for them

- more controllable size and output
- slightly faster user bundling
- faster browser loading if somebody will use it
- better debugging in development mode without sourcemaps (less webpack stuff)
- hidden internals which may cause breaking changes if user decide to use them

Size snapshot shows also treeshakability for esm bundles so I can apply
more optimisation which will be visible to us.

/cc @nkbt @chenglou


package.json
```diff
@@ -2,20 +2,21 @@
   "name": "react-motion",
   "version": "0.5.2",
   "description": "A spring that solves your animation problems.",
-  "main": "lib/react-motion.js",
+  "main": "dist/react-motion.cjs.js",
+  "module": "dist/react-motion.esm.js",
```
nkbt: should we use `dist/react-motion.jsm` or does not really make much sense for non-nodejs library?

package.json
```diff
@@ -56,10 +57,9 @@
   },
   "scripts": {
     "start": "node server.js",
-    "build:lib": "cross-env NODE_ENV=cjs babel src --out-dir lib",
```
nkbt: that would mean a major release then, it certainly is a breaking change

nkbt: I am not sure this change worth it

TrySound:
@nkbt Could you merge, so I could start working on further improvements?

PR #550, opened 2018-08-10
Improve babel config

TrySound:
- enable loose mode for env and class properties
- reuse helpers from babel runtime instead of duplicating them

This change saved us 2kb minified.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,11 +1,11 @@
 {
   "presets": [
-    ["@babel/env", { "modules": false }],
+    ["@babel/env", { "modules": false, "loose": true }],
     "@babel/flow",
     "@babel/react"
   ],
   "plugins": [
-    "@babel/proposal-class-properties"
+    ["@babel/proposal-class-properties", { "loose": true }]
   ],
   "env": {
     "test": {
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,8 @@
     "@babel/react"
   ],
   "plugins": [
-    ["@babel/proposal-class-properties", { "loose": true }]
+    ["@babel/proposal-class-properties", { "loose": true }],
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -92,6 +93,7 @@
   ],
   "license": "MIT",
   "dependencies": {
+    "@babel/runtime": "7.0.0-rc.1",
```
nkbt: I haven’t used babel runtime since docs are very loose and end effect on final app is not clear.

The fact lib bundle gets smaller is good of course. But does it mean that babel runtime will be (has to be) added to an app?
When you have build in place that is not a big deal, but in case of static sites using minified prebuilt versions of libs it means no tree shaking. Full babel runtime is huge :(

Please correct me if I am wrong here (I wish I am)

.babelrc
```diff
@@ -1,11 +1,11 @@
 {
   "presets": [
-    ["@babel/env", { "modules": false }],
+    ["@babel/env", { "modules": false, "loose": true }],
```
nkbt: :+1: for loose

package.json
TrySound: It's direct dependency, so no affect to user. `@babel/plugin-transform-runtime` imports only necessary stuff. Here's what imported now
```js
import _extends from '@babel/runtime/helpers/esm/extends';
import _inheritsLoose from '@babel/runtime/helpers/esm/inheritsLoose';
```

it's exactly the same what `rollup-plugin-babel` inlines to the bundle without runtime plugin, so their's not gain or loss in size. The idea is that these helpers may be reused across many packages.

Actually the new babel runtime package is small. It was decoupled from core-js version. So even less stuff in user node_modules.

TrySound:
@nkbt rebased and added a few tweaks

package.json
nkbt: Thanks for the explanation!

PR #551, opened 2018-08-12
Export SpringHelperConfig type

TrySound:
This type is used by [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd/blob/master/src/view/moveable/moveable.jsx#L3).

Flat bundles and single types reexport will forbid using internals so we
need to expose some types like this one. If something else will be
required we could make feature released after 0.6.

### DIFF
--- a/src/react-motion.js
+++ b/src/react-motion.js
@@ -1,4 +1,7 @@
 /* @flow */
+
+export type { SpringHelperConfig } from './Types';
+
 export { default as Motion } from './Motion';
 export { default as StaggeredMotion } from './StaggeredMotion';
 export { default as TransitionMotion } from './TransitionMotion';


src/react-motion.js
```diff
@@ -1,4 +1,7 @@
 /* @flow */
+
+export type { SpringHelperConfig } from './Types';
```
nkbt: Isn't it better to export all types? Sorry, I'm not using flow/ts, so cannot do much feedback on how it may affect downstream

TrySound: Some of types are internal. I think types are also a part of api. api with all possible types will be too verbose. I would export types with a request. This is the first one. Maybe the last one.

PR #553, opened 2018-08-28
Fix scale style in demo7

jsm1003:
The online demo7 doesn't work as  expected, I fix it.

### DIFF
--- a/demos/demo7-water-ripples/Demo.jsx
+++ b/demos/demo7-water-ripples/Demo.jsx
@@ -38,7 +38,7 @@ export default class Demo extends React.Component {
       key: now,
       style: {
         opacity: spring(1),
-        scale: spring(0),
+        scale: 1,
         x: spring(mouseX),
         y: spring(mouseY),
       }


nkbt:
I've just checked demo7 and it works as expected, what is the problem you think there is?

jsm1003:
I created a [GIF,](https://github.com/jsm1003/asdf) you can see when I open the demo7 from my computer, and there were no water ripples

nkbt:
This is how it works for me. I checked in opera and under windows 10 too. Same way everywhere

ripples

I do not know why you are seeing nothing. Maybe it is opened devtools (so chrome throttles something), maybe some browser plugin.

PS: I recommend [Licecap](https://www.cockos.com/licecap/) to do gif captures. Very cool thing

jsm1003:
@nkbt 
 Watch this Gif , you can see demo-ball element add an attribute , scale : 0

react-motion-demo2

my chrome version : 68.0.3440.106

I checked this URL from my classmate's computer, it works well ,It seems that only the chrome of my computer has this problem.

nkbt:
I am not sure I want to merge fix that fixes something broken on the only machine... Would better wait if someone else can reproduce it too.

PR #555, opened 2018-10-09
React-motion: Removed phantomjs dependency

ossdev07:
PhantomJS is an unmaintained project removed phantomjs dependency from package.json and
Added support for Karma-chrome-launcher.

Signed-off-by: ossdev07 <ossdev@puresoftware.com>


nkbt:
Thanks, looks like a minimal change and all tests still work well

PR #556, opened 2018-10-18
Add live demo to README.md

ipiranhaa:


README.md
```diff
@@ -5,6 +5,8 @@
 [![Bower version](https://badge.fury.io/bo/react-motion.svg)](http://badge.fury.io/bo/react-motion)
 [![react-motion channel on discord](https://img.shields.io/badge/discord-react--motion%40reactiflux-738bd7.svg?style=flat)](https://discordapp.com/invite/0ZcbPKXt5bYzmcI0)
 
+![Live Demo](https://media.giphy.com/media/mz8Y3zn2ZPMMA95BIp/giphy.gif)
```
nkbt: It would be better to add gif to the repo and reference here instead of `giphy`
Also I suggest using `licecap` as it can do much better fps (up to 60) than this one.

Issue #569, opened 2019-08-12
componentWillReceiveProps warning with React 16.9


alexbensimon:
Hello,

After updating `react` and `react-dom` to v`16.9.0`, this warning appears: 

```
react-dom.development.js:11494 Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-async-component-lifecycle-hooks for details.

* Move data fetching code or side effects to componentDidUpdate.
* If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state
* Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

Please update the following components: Motion
```

According to https://reactjs.org/blog/2019/08/08/react-v16.9.0.html#renaming-unsafe-lifecycle-methods, the `componentWillReceiveProps` is now deprecated and will be removed from the API on the next major version.

juicylevel:
2020 year..

nkbt:
Yeah @juicylevel, I wish I could drop all the things of my life now (as in work, surfing, divorce, girlfriend) and do all the opensource...

On a side note: I've added a reminder to publish new version of react-motion which should have this fixed (at least temporarily). I'll let everyone know when it is published. You can help by teting current master branch in your real-world project and see if anything breaks.

jfroffice:
Any update on this ?

comp615:
When is the reminder to publish set for? Are there any other maintainers who might be able to help you out and publish this in the meantime? I know it's annoying to have people nagging, but getting the newer React support updated would be super helpful! Thank you!

fabiomcosta:
Ah interesting... so sounds like the change is already there and we just need a new version published to `npm`?
If the maintainers give me the permissions I'll happily do that.

andresmatasuarez:
yes, please!

LiteSoul:
...and now it's 2021 🥺

RaddishIoW:
Someone has forked react-motion to fix this issue: [see this thread](https://github.com/chenglou/react-motion/issues/604)

onionhammer:
Perhaps other projects should reference the fork

RaddishIoW:
[...]

PR #575, opened 2019-11-08
rename react lifecycle methods to prefix "UNSAFE_"

theopak:
hi @chenglou and @nkbt! what do you think of potentially shipping this quick rename? it comes from `$ npx react-codemod rename-unsafe-lifecycles` and would make the developer experience a little better for consumers that are on react v16.9, like we are at @Frameio

for anyone just catching up with this situation — please note that this rename fixes the console warnings introduced in react@16.9 such that code will run in react@17, but this rename _doesn't_ address the underlying issue

_edit:_ i pushed a config change just to get the build to pass, so please lmk if there's a different way that would work better here


theopak:
_technically_ fixes #569

theopak:
@nkbt awesome! i don't have merge permission though, can you merge this? 

yes, the eslint change was just to get the build to pass

nkbt:
I've tested this PR and also check if built version works ok in one of my projects and everything seems to work well without React warnings. Merging in.

nkbt:
Cannot publish until https://github.com/chenglou/react-motion/issues/577

petrkrejcik:
Hi @nkbt, will you make a release with this change?

PR #579, opened 2019-12-06
Bump version to v0.6.0

davidmh:
I'm doing a minor update along with a stricter react peer dependency to
match the deprecation of the unsafe lifecycle methods.


alexweber:
Can we get this merged?  Seems this will benefit anyone using @nivo stuff for charts...

PR #603, opened 2021-01-12
customize the build for KA

k4b7:
We only include those things that we using in webapp.  In the future we may want to include `TransitionMotion` in order to replace `react-transition-group`, but we haven't made that decision yet.

I ran `npm run prepublishOnly` to generate build files.


k4b7:
I meant to make this a pull request against my fork.  Sorry for the noise.

PR #619, opened 2021-10-06
Chore: add src to npmignore

DmitryMarkov:


.npmignore
```diff
@@ -1,9 +1,14 @@
...
 
+/bin/
+/chrome.json
+/renovate.json
+
 /bower_components/
 
 /coverage/
 /demos/
+/src/
```
nkbt, line 11: On a general note it is a very bad idea. This means consumers of the library cannot build and do all the optimisations they potentially could. 
If you ever publish any library please always include sources in the final published package

